PR #2, opened 2019-04-18
Abstraction of Original Game Files access

ataulien:
This class provides easy access to the original game files for all platforms.
                                                                             
Since Gothic was developed on Windows-Systems, which uses a case-insensitive 
file system, there are major differences in file naming between releases.    
Sometimes a folder is called `Data`, `DATA` or `data`, you get the idea.     
                                                                             
This is troublesome for UNIX-systems which generally use case-sensitive      
file systems.                                                                
                                                                             
Another reason this class exists is so we can abstract away the paths to     
certain files used by the engine.                                            
                                                                             
Generally, all paths you get back are in the correct case for your system.

### DIFF
--- a/src/REGothEngine.cpp
+++ b/src/REGothEngine.cpp
@@ -18,25 +18,21 @@ REGothEngine::~REGothEngine()
   shutdown();
 }
 
-bs::Vector<bs::Path> REGothEngine::getVdfsPackagesToLoad()
-{
-  return gOriginalGameFiles().allVdfsPackages();
-}
-
 void REGothEngine::loadOriginalGamePackages(const bs::String& argv0, const bs::Path& gameDirectory)
 {
+  OriginalGameFiles files = OriginalGameFiles(gameDirectory);
+
   gVirtualFileSystem().setPathToEngineExecutable(argv0);
-  gOriginalGameFiles().setOriginalFilesRoot(gameDirectory);
 
   bs::gDebug().logDebug("[VDFS] Indexing packages: ");
 
-  for (auto p : getVdfsPackagesToLoad())
+  for (auto p : files.allVdfsPackages())
   {
     bs::gDebug().logDebug("[VDFS]  - " + p.getFilename());
     gVirtualFileSystem().loadPackage(p);
   }
 
-  gVirtualFileSystem().mountDirectory(gOriginalGameFiles().vdfsFileEntryPoint());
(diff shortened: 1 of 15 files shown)


Joshin95:
LGTM 👍

mworchel:
I am not particularly familiar with the workings of the original engine or planned features but why did you make the OriginalGameFile class a singleton instead of e.g. the REGothEngine class having an instance of it as a member? Is global access to it ever needed?

ataulien:
There are some other modules which will need access to the original files. However, now that I think of it, most of these should be handled through the virtual file system rather.
@frabert Do you know whether mounting the file system works via physfs? 

 -

frabert:
Yes, PhysFS natively supports mounting _both_ archives and file-system directories[1]. This was indeed one of my main interests, in that we could, for example, unpack a VDF file on the disk (for modding uses, or maybe debugging) and fire up the engine without repacking it

[1] https://www.icculus.org/physfs/docs/html/physfs_8h.html#a8eb320e9af03dcdb4c05bbff3ea604d4

src/REGothEngine.cpp
```diff
@@ -17,48 +18,23 @@ REGothEngine::~REGothEngine()
   shutdown();
 }
 
-bs::Vector<bs::String> REGothEngine::getVdfsPackagesToLoad(const bs::Path& dataDirectory)
+bs::Vector<bs::Path> REGothEngine::getVdfsPackagesToLoad(const bs::Path& dataDirectory)
```
mworchel: Sorry deleted my previous comment. The parameter `dataDirectory` seems to be obsolete now.

ataulien: Removed it in 59b0320.

ataulien:
How's this?

`OriginalGameFiles` is no longer a singleton. I rather mounted those files into the VDFS, like the original game is doing as well, so everything can use the VDFS to get its data.

src/REGothEngine.hpp
```diff
@@ -7,6 +7,8 @@
 
 namespace REGoth
 {
+  class OriginalGameFiles;
```
mworchel: The forward declaration is not necessary, is it?

mworchel:
Looks a lot cleaner now, as far as I can tell from my limited point of view.

src/REGothEngine.hpp
ataulien: Ah, I put it there before I removed the method in which you list all vdfs packages. Good catch, I'll remove it!

PR #3, opened 2019-04-20
Node attachments

ataulien:
This PR adds the `VisualInteractiveObject` for use with MOBs. It also refactors the `VisualCharacter` so that much code could be put into a generic `VisualSkeletalAnimation` which allows much code to be shared between VisualInteractiveObject` and VisualCharacter`.

It also adds loading of interactive objects from Zen for some simple objects. However, the objects orientations might not be correct. This is because bs:f does not seem to move objects attached to bones to their correct location when no animation is playing...

### DIFF
--- a/src/components/VisualSkeletalAnimation.hpp
+++ b/src/components/VisualSkeletalAnimation.hpp
@@ -162,6 +162,12 @@ namespace REGoth
 
   protected:
     /**
+     * @return A list of animations to try playing after initialization or
+     *         if no animation is playing.
+     */
+    virtual bs::Vector<bs::String> listPossibleDefaultAnimations() const;
+
+    /**
      * Called when a animation event was triggered
      */
     virtual void onAnimationEvent(const bs::HAnimationClip& clip, bs::String string);
--- a/src/components/VisualSkeletalAnimation.hpp
+++ b/src/components/VisualSkeletalAnimation.hpp
@@ -1,0 +1,271 @@
+/** \file
+ */
+
+#pragma once
+
+#include <BsPrerequisites.h>
+#include <BsZenLib/ZenResources.hpp>
+#include <Scene/BsComponent.h>
+
+namespace REGoth
+{
+  class RTTI_VisualSkeletalAnimation;
+  class NodeVisuals;
+  using HNodeVisuals = bs::GameObjectHandle<NodeVisuals>;
+
+  /**
+   * Component for rendering an object animated using skeletal animation (Player, NPC, Monster).
+   *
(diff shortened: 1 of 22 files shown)


ataulien:
Rebased onto master and force-pushed to resolve merge conflicts.

src/components/VisualSkeletalAnimation.hpp
```diff
@@ -0,0 +1,270 @@
+/** \file
```
mworchel: Default doxygen comment leftover?

src/original-content/VirtualFileSystem.cpp
```diff
@@ -1,6 +1,6 @@
 #include "VirtualFileSystem.hpp"
-#include <Error/BsException.h>
 #include <FileSystem/BsFileSystem.h>
+#include <excepction/Throw.hpp>
```
mworchel: We should fix that typo soon, before everything get's too big. But that's something for another PR.

mworchel:
Does not compile. You forgot to add and commit main_MobViewer.cpp.

src/original-content/VirtualFileSystem.cpp
ataulien: Oh, that absolutely went past me. We should definitely fix this!

src/components/VisualSkeletalAnimation.hpp
ataulien: This is needed for the file to be recognized by Doxygen at all. The documentation can be found at the Class then.

ataulien:
That's unfortunate. Usually I just do a `git add .` but this time I had some other files I did not want to commit. You can just remove that sourcefile from the build for now, using the WaynetTester should also work.

mworchel:
Still doesn't compile:
`error C2039: 'ShaderKind': is not a member of 'BsZenLib'`
Did you forget to check in the file where the Enum `BsZenLib::ShaderKind` is defined?

ataulien:
That might be true. I can't update the submodule right now. Can you try to update it manually?

src/original-content/VirtualFileSystem.cpp
mworchel: I can do a PR for that. It's currently only 13 files which have to be fixed (+ the CMakeLists.txt where I forgot to include it in my PR #4).

mworchel:
Compiles and runs after using the master branch of `BsZenLib`. Tested with the `WorldViewer` and `WaynetTester`. Anything specific to look for?

ataulien:
Chests and some other interactive objects should be in the world now.

PR #4, opened 2019-04-20
Make library header files show up in IDEs

mworchel:
When using CMake to generate a solution for an IDE like Visual Studio, header files files must be passed to the `add_library` or `add_executable`  commands in order to show up in the IDE. Although header files themselves have not to be compiled, seeing them in the IDE definitely improves the user experience.

This pull request adds the header files to the list of the sources for the libraries `samples-common` and `REGothEngine`.


ataulien:
Looks good, thank you!

PR #5, opened 2019-04-21
Add component for drawing text labels anchored at 3d world positions

mworchel:
This PR adds a component which can be used to draw text labels anchored at specific locations in 3d world space. It mainly serves debug purposes i. e. displaying some text related to an object above the object when it is visible. The component can be added to any scene object that possesses a valid CGUIWidget component.


src/common/BsAnchoredTextLabels.cpp
```diff
@@ -0,0 +1,52 @@
+#include "BsAnchoredTextLabels.h"
+#include "Components/BsCCamera.h"
+#include "GUI/BsCGUIWidget.h"
+#include "GUI/BsGUILabel.h"
+#include "GUI/BsGUIPanel.h"
+#include "Math/BsMath.h"
+#include "Scene/BsSceneObject.h"
+
+namespace bs
```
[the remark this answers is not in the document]
mworchel: I see, didn't know this. I used the given components as a guidelines that's why I didn't use <> for the includes and didn't throw when the component isn't found.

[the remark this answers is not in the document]
ataulien: Take a look into `src/components/` for the "real" examples written for regoth.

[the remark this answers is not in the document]
ataulien: Generally, I like to use " only for stuff that is directly reachable from the current directory.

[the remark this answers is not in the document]
mworchel: Thanks for clearing that up. I was asking myself what your convention regarding <> and "" is.

src/common/BsAnchoredTextLabels.cpp
```diff
@@ -0,0 +1,52 @@
+#include "BsAnchoredTextLabels.h"
+#include "Components/BsCCamera.h"
```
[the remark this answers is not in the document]
mworchel: Resolved in 49accda

src/common/BsAnchoredTextLabels.cpp
```diff
@@ -0,0 +1,52 @@
...
+namespace bs
+{
+  AnchoredTextLabels::AnchoredTextLabels(const HSceneObject& parent)
+      : Component(parent)
+  {
+    // Set a name for the component, so we can find it later if needed
+    setName("AnchoredTextLabels");
+
+    // A GUI widget component is required for drawing the labels
+    mGui = parent->getComponent<CGUIWidget>();
```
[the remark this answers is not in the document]
mworchel: Resolved in 49accda.

src/common/BsAnchoredTextLabels.cpp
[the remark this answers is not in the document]
mworchel: Resolved in 49accda

src/components/AnchoredTextLabels.cpp
```diff
@@ -0,0 +1,61 @@
...
+    }
+  }
+
+  void AnchoredTextLabels::addLabel(const bs::Vector3& anchorPosition, const bs::HString& text)
+  {
+    bs::GUILabel* label = bs::GUILabel::create(text);
+    mLabels.emplace_back(AnchoredLabel{anchorPosition, label});
+
+    bs::GUIPanel* mainPanel = mGui->getPanel();
+    mainPanel->addElement(label);
```
ataulien: Should also remove those elements again in the destructor.

src/components/AnchoredTextLabels.cpp
```diff
@@ -0,0 +1,61 @@
...
+  }
+
+  void AnchoredTextLabels::update()
+  {
+    auto camera                          = mGui->getCamera();
+    const bs::Transform& cameraTransform = camera->getTransform();
+    const bs::Vector3& cameraPosition    = cameraTransform.getPosition();
+    const bs::Vector3& cameraForward     = cameraTransform.getForward();
```
ataulien: Please add a linebreak here, looks too crowded otherwise. Also a linebreak after initializing `camera` would be nice.

src/components/AnchoredTextLabels.cpp
```diff
@@ -0,0 +1,61 @@
...
+    for (const auto& anchoredLabel : mLabels)
+    {
+      // Only show labels with an anchor point in front of the camera
+      if ((anchoredLabel.mWorldPosition - cameraPosition).dot(cameraForward) > 0.f)
```
ataulien: Could be extracted into a function `isPointInFrontOfCamera` or so.

src/components/AnchoredTextLabels.cpp
```diff
@@ -0,0 +1,61 @@
...
+namespace REGoth
+{
+  AnchoredTextLabels::AnchoredTextLabels(const bs::HSceneObject& parent)
+      : Component(parent)
+  {
+    // Set a name for the component, so we can find it later if needed
+    setName("AnchoredTextLabels");
+
+    // A GUI widget component is required for drawing the labels
+    mGui = parent->getComponent<bs::CGUIWidget>();
```
ataulien: Why are you taking the parents component? Couldn't we just add the  `AnchoredTextLabels` component to the same scene object?

ataulien: Or maybe even allow passing a handle to CGUI via the constructor. Then we could have multiple instances of the component managed by whatever needs it... Or should we just search up the parent chain?

src/components/AnchoredTextLabels.cpp
mworchel: Good catch. When I want to remove the labels in the destructor, I have to get the `mainPanel` again from `mGui`. Is there a guarantee that the `CGUIWidget` component is destroyed *after* mine if it has been added to the scene object before mine (e. g. in the case of the application being shut down)? If there is not, I should probably check if `mGui` is already destroyed before trying to remove the labels.

mworchel: An alternative would be using the `onDestroyed()` method. Would probably be the best alternative as the bs::Component documentation says 
> Avoid putting logic in constructors or destructors.

src/components/AnchoredTextLabels.cpp
mworchel: Fixed in 2a3cd51.

src/components/AnchoredTextLabels.cpp
ataulien: `onDestroyed` seems to be the correct way to go. If the `CGUI` is destroyed, you won't need to remove any labels anyways. You should still be checking whether the handle to `mGui` is valid to not cause an exception when accessing the destroyed object.

src/components/AnchoredTextLabels.cpp
mworchel: It is supposed to be added to a scene object which has a `CGUIWidget` component already. I mean you could definitely loosen the constraint and pass the required component in the constructor. This is the way I would favor, as it still allows the add the component to a "GUI scene object" or just any other scene object, if you e. g. have a scene object which manages all debug related drawing or something like that.

src/components/AnchoredTextLabels.cpp
mworchel: I would expect such a function to take a camera and a position as input. Unfortunately that wouldn't allow me to get the camera position and forward direction only once beforehand but I would have to do it for every label. I mean it's some kind of premature optimization but I thought this way would be the most future proof one.

By the way, thinking of this line I had the idea that it would probably be helpful to specify something like a maximum distance. Like "only draw these labels if their anchor points are closer than *x* to the camera". Could be easily done as the dot product gives us the distance (assuming `cameraForward` is normalized). What do you thinkg about it?

src/components/AnchoredTextLabels.cpp
ataulien: > It is supposed to be added to a scene object which has a CGUIWidget component already. 

Then you don't want `parent`, but rather `SO()`, which is the same thing but it's the common notation for accessing a components scene object. There is also `SO()->parent()` or something which threw me off. I thought you were searching the scene object which is the parent to the scene object the component is attached to.

src/components/AnchoredTextLabels.cpp
ataulien: I don't remember how Gothic did it, but right now all the text would be visible through walls, so adding a max-distance would be the easiest solution to not clutter the screen now.

As for optimization, just pass a reference to the cameras transform. The compiler should be smart enough to optimize the calls to `getForward()` etc. It would make adding the distance check cleaner as well IMHO.

src/components/AnchoredTextLabels.cpp
mworchel: Fixed in 3dbd4fd.

src/components/AnchoredTextLabels.cpp
mworchel: `parent` is the scene object which owns the component. I see why you were confused though. I will add the `CGUIWidget` component as argument to the constructor. This should make things more clear.

src/components/AnchoredTextLabels.cpp
mworchel: I added the distance check and separate function in 6889ece. We could potentially hide text labels if their anchor point is behind world geometry by looking at the depth buffer of the attached camera. But I don't know how and if that is possible and it would probably be an overkill ATM.

src/components/AnchoredTextLabels.cpp
mworchel: Done in 1b6da35.

src/components/AnchoredTextLabels.h
```diff
@@ -32,9 +37,13 @@ namespace REGoth
       bs::GUILabel* mLabel;
     };
 
+    bool isPointWithinForwardDrawDistance(const bs::Transform& cameraTransform, const bs::Vector3& point);
```
ataulien: That can be `const`. Otherwise it looks fine!

mworchel: Good point. Done in 72fbf32.

PR #6, opened 2019-04-22
REGoth Engine Resources (Shaders, etc) loading

ataulien:
This PR adds modules which can find a `content`-directory where resources needed by the engine can be stored and load them. Also it refines the resource management inside the base `REGothEngine`-class and should fix the "black textures"-bug for good.

The main change however is the new shader applied to most of the worlds geometry. It has less fancy features than the default one (ie. no PBR), but supports vertex lighting and masked textures.

### DIFF
--- a/content/shaders/World.bsl
+++ b/content/shaders/World.bsl
@@ -1,0 +1,169 @@
+#include "$ENGINE$\GBufferOutput.bslinc"
+#include "$ENGINE$\PerCameraData.bslinc"
+#include "$ENGINE$\PerObjectData.bslinc"
+#include "$ENGINE$\VertexInput.bslinc"
+
+shader Surface
+{
+  mixin GBufferOutput;
+  mixin PerCameraData;
+  mixin PerObjectData;
+  mixin VertexInput;
+
+  code
+  {
+    [alias(gAlbedoTex)] SamplerState gAlbedoSamp;
+
+    Texture2D gAlbedoTex = white;
+
+    struct ZenVertexInput
+    {
+      #if POSITION_2D
+        float2 position : POSITION;
+      #else
+        float3 position : POSITION;
+      #endif
+
+      float2 uv0 : TEXCOORD0;
+
+      #if LIGHTING_DATA
+        float3 normal : NORMAL; // Note: Half-precision could be used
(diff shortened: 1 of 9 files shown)


content/shaders/World.bsl
```diff
@@ -0,0 +1,169 @@
...
+
+  code
+  {
+    [alias(gAlbedoTex)] SamplerState gAlbedoSamp;
+
+    Texture2D gAlbedoTex = white;
+
+    struct ZenVertexInput
+    {
+      #if POSITION_2D
```
mworchel: Don't you need a `variation` to do this or is it not yet used?

ataulien: I don't know, this was copied from the original vertex input structure. I haven't read that deep into the shader language yet. However, I've seen some variations being defined in one of the included headers.

src/REGothEngine.cpp
```diff
@@ -181,12 +222,11 @@ int ::REGoth::main(REGothEngine& regoth, int argc, char** argv)
 
   regoth.loadCachedResourceManifests();
 
+  regoth.setShaders();
   regoth.setupInput();
   regoth.setupMainCamera();
   regoth.setupScene();
 
-  regoth.saveCachedResourceManifests();
```
mworchel: You forgot to remove the same line in `src/main.cpp`, so the `REGoth` target doesn't compile.

mworchel:
Add targets except for the `REGoth` one compile but no textures are shown (everything is black). I deleted the cache folder and even deleted the whole build directory and recompiled. Tested with the `WaynetTester`.

ataulien:
I can reproduce the black textures. Damn, I was hoping this was an issue with the standard shader and/or resource manifest. I guess I'll have to dig deeper into why that happens.

There is also a huge slowdown caused by saving the resource manifest so often, so I need to get rid of that too.

ataulien:
Black Textures should be resolved now. @mworchel can you try again, please?

mworchel:
> 
> 
> Black Textures should be resolved now. @mworchel can you try again, please?

I tried it and the issue seems to be resolved. No more black textures.

ataulien:
Awesome! I'll merge it then.

PR #7, opened 2019-04-22
Draw waypoint names above waypoints

mworchel:
This PR adds drawing of `Waypoint` names to the debug draw of the `Waynet`. A name appears as text label above their corresponding waypoint in 3d world space.


ataulien:
Tested and works!

PR #16, opened 2019-04-28
Feature/gameclock

KirmesBude:
Implements the GameClock component.
Not currently integrated to the world at all ;D.
It also keeps track of the elapsed real-time seconds for fun. Maybe this is useful somehow or can be used to display playtime somewhere.

The functionality is derived from the three Wld_* externals that do something with the time/clock.
If there are more features needed, let me know.

### DIFF
--- a/src/RTTI/RTTI_GameClock.hpp
+++ b/src/RTTI/RTTI_GameClock.hpp
@@ -1,0 +1,40 @@
+#pragma once
+
+#include "RTTI_TypeIDs.hpp"
+#include <BsCorePrerequisites.h>
+#include <Private/RTTI/BsGameObjectRTTI.h>  // Says private, but bs:f uses this too in their RTTIs
+#include <Reflection/BsRTTIType.h>
+#include <components/GameClock.hpp>
+
+namespace REGoth
+{
+  class RTTI_GameClock
+      : public bs::RTTIType<GameClock, bs::Component, RTTI_GameClock>
+  {
+    BS_BEGIN_RTTI_MEMBERS
+    // TODO: Fill RTTI Members
+    BS_END_RTTI_MEMBERS
+
+  public:
+    RTTI_GameClock()
+    {
+    }
+
+    bs::SPtr<bs::IReflectable> newRTTIObject() override
+    {
+      return bs::GameObjectRTTI::createGameObject<GameClock>();
+    }
+
+    const bs::String& getRTTIName() override
+    {
+      static bs::String name = "GameClock";
+      return name;
+    }
+
+    bs::UINT32 getRTTIId() override
+    {
+      return TID_REGOTH_GameClock;
+    }
+  };
+
+}  // namespace REGoth
--- a/src/RTTI/RTTI_GameClock.hpp
+++ b/src/RTTI/RTTI_GameClock.hpp
@@ -21,21 +21,7 @@ namespace REGoth
     {
     }
 
-    bs::SPtr<bs::IReflectable> newRTTIObject() override
-    {
-      return bs::GameObjectRTTI::createGameObject<GameClock>();
-    }
-
-    const bs::String& getRTTIName() override
-    {
-      static bs::String name = "GameClock";
-      return name;
-    }
-
-    bs::UINT32 getRTTIId() override
-    {
-      return TID_REGOTH_GameClock;
-    }
+    REGOTH_IMPLEMENT_RTTI_CLASS_FOR_COMPONENT(GameClock)
   };
 
 }  // namespace REGoth
--- a/src/RTTI/RTTI_GameClock.hpp
+++ b/src/RTTI/RTTI_GameClock.hpp
@@ -13,7 +13,7 @@ namespace REGoth
   {
     BS_BEGIN_RTTI_MEMBERS
       BS_RTTI_MEMBER_PLAIN(mElapsedSeconds, 0)
-      BS_RTTI_MEMBER_PLAIN(mElapsedIngameSeconds, 0)
+      BS_RTTI_MEMBER_PLAIN(mElapsedIngameSeconds, 1)
     BS_END_RTTI_MEMBERS
 
   public:
--- a/src/RTTI/RTTI_GameClock.hpp
+++ b/src/RTTI/RTTI_GameClock.hpp
@@ -12,7 +12,8 @@ namespace REGoth
       : public bs::RTTIType<GameClock, bs::Component, RTTI_GameClock>
   {
     BS_BEGIN_RTTI_MEMBERS
-    // TODO: Fill RTTI Members
+      BS_RTTI_MEMBER_PLAIN(mElapsedSeconds, 0)
+      BS_RTTI_MEMBER_PLAIN(mElapsedIngameSeconds, 0)
     BS_END_RTTI_MEMBERS
 
   public:
--- a/src/RTTI/RTTI_TypeIDs.hpp
+++ b/src/RTTI/RTTI_TypeIDs.hpp
@@ -42,5 +42,6 @@ namespace REGoth
     TID_REGOTH_ScriptSymbolUnsupported      = 600034,
     TID_REGOTH_ScriptClassTemplates         = 600035,
     TID_REGOTH_ScriptObjectMapping          = 600036,
+	TID_REGOTH_GameClock					= 600037,
   };
 }  // namespace REGoth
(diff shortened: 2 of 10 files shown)


src/components/GameClock.hpp
```diff
@@ -0,0 +1,86 @@
...
+
+    /**
+     * @return Current ingame day.
+     */
+    bs::INT32 getDay() const;
+
+    /**
+     * @return Current hour of time of day (hh:mm).
+     */
+    bs::UINT8 getHour() const;
```
ataulien: Shouldn't those be INT32 as well?

src/RTTI/RTTI_GameClock.hpp
```diff
@@ -0,0 +1,40 @@
...
+#include <Reflection/BsRTTIType.h>
+#include <components/GameClock.hpp>
+
+namespace REGoth
+{
+  class RTTI_GameClock
+      : public bs::RTTIType<GameClock, bs::Component, RTTI_GameClock>
+  {
+    BS_BEGIN_RTTI_MEMBERS
+    // TODO: Fill RTTI Members
```
ataulien: Should be eas to add since the class only stores two floats. If you need examples, check the serialization branch

src/components/GameClock.cpp
```diff
@@ -0,0 +1,106 @@
...
+  constexpr float SECONDS_IN_A_MINUTE = 60;
+  constexpr float SECONDS_IN_AN_HOUR  = 60 * SECONDS_IN_A_MINUTE;
+  constexpr float SECONDS_IN_A_DAY    = 24 * SECONDS_IN_AN_HOUR;
+  constexpr float CLOCK_SPEED_FACTOR  = SECONDS_IN_A_DAY/6000; // see https://forum.worldofplayers.de/forum/threads/396326-Tipp-Tageweise-springen-%28auch-zur%C3%BCck%29?p=6231841&viewfull=1#post6231841
+
+  GameClock::GameClock(const bs::HSceneObject& parent)
+    : bs::Component(parent)
+  { }
+
+  void GameClock::update()
```
ataulien: There is also `fixedUpdate` which should be better here. Otherwise we may lose some time or weird stuff happens on long frames.

src/RTTI/RTTI_GameClock.hpp
```diff
@@ -0,0 +1,41 @@
...
+#include <components/GameClock.hpp>
+
+namespace REGoth
+{
+  class RTTI_GameClock
+      : public bs::RTTIType<GameClock, bs::Component, RTTI_GameClock>
+  {
+    BS_BEGIN_RTTI_MEMBERS
+      BS_RTTI_MEMBER_PLAIN(mElapsedSeconds, 0)
+      BS_RTTI_MEMBER_PLAIN(mElapsedIngameSeconds, 0)
```
ataulien: The last parameter is supposed to be a unique number. Should be set to 1 here.

KirmesBude: Woops.
Fixed it

KirmesBude:
How to proceed with this PR?
Should we try to address the remaining TODO comments?
I intentionally did not "link" the externals or integrated the component into the world, because that would need to be reworked following you pull request. Shall this PR be integrated first and we worry about the other stuff later or in your PR?
Is there anything missing from the GameClock component for now?

ataulien:
I agree that you shouldn't integrate it right now. However, since the serialization branch should be done, apart from some minor issues, we could either wait until it is merged or you can just rebase your branch onto the serialization branch so you can start the integration.

ataulien:
About missing features: I noticed there was no way to set the current day other than "abusing" `setTime`. I'm not sure whether old REGoth had one, but I feel this is something you should be able to set.

src/RTTI/RTTI_TypeIDs.hpp
```diff
@@ -5,12 +5,43 @@ namespace REGoth
...
+    TID_REGOTH_ScriptSymbolString           = 600028,
+    TID_REGOTH_ScriptSymbolClass            = 600029,
+    TID_REGOTH_ScriptSymbolScriptFunction   = 600030,
+    TID_REGOTH_ScriptSymbolExternalFunction = 600031,
+    TID_REGOTH_ScriptSymbolPrototype        = 600032,
+    TID_REGOTH_ScriptSymbolInstance         = 600033,
+    TID_REGOTH_ScriptSymbolUnsupported      = 600034,
+    TID_REGOTH_ScriptClassTemplates         = 600035,
+    TID_REGOTH_ScriptObjectMapping          = 600036,
+	TID_REGOTH_GameClock					= 600037,
```
mworchel: Used tabs instead of spaces?

KirmesBude: Ohh. I have VSCodium set up correctly, but this change I did with BeyondCompare.
Need to change the settings there, too.

KirmesBude:
> About missing features: I noticed there was no way to set the current day other than "abusing" `setTime`. I'm not sure whether old REGoth had one, but I feel this is something you should be able to set.
> 
> You could also see if you can find out how sleeping is handled. If the scripts define the times to sleep until it is fine, but if is using an enum of some sort a `advanceToNext`-function taking values like `Evening`, `Midnight`, etc would be nice to have.

As far as I can tell, they call their internal Sleep function with 8, 12, 20 or 0. There they check with WLD_isTime if we need to skip a day (add 24) or stay on this day. With the result they call WLD_setTime. 
 
Though now that I have the externals also registered I wonder where these Gothic specific things should be handled. Maybe that should be at the externals (DaedalusVMForGameWorld) and the GameClock component could be streamlined a lot more?

src/components/GameClock.hpp
```diff
@@ -66,10 +66,14 @@ namespace REGoth
      */
     void setTime(bs::INT32 hour, bs::INT32 min);
 
+    void setDay(bs::UINT32 day);
+
   private:
     float mElapsedSeconds       = 0.0;
```
ataulien: Should rather be `0.0f` so it doesn't do an implicit cast from double.

src/components/GameWorld.cpp
```diff
@@ -58,7 +58,9 @@ namespace REGoth
 
     onImportedZEN();
 
-    mGameClock = SO()->addComponent<GameClock>();
```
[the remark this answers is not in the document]
KirmesBude: I was not quite sure how to do it anyways :D.
How will changing worlds work? Maybe having the GameClock here does not make too much sense? Not sure if Gothic keeps track of time between worlds.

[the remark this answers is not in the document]
markusobi: > I was not quite sure how to do it anyways :D.
> How will changing worlds work? Maybe having the GameClock here does not make too much sense? Not sure if Gothic keeps track of time between worlds.

The game time (`GameClock`) is bound to the current session. That means it must be persistent across world switch. The following things are transferred when switching worlds:
 - the player entity (should not be saved to the world-savegame)
 - the game time (should not be saved to the world-savegame)
 - global script variables (they reflect i.e. quest progress)

The GameClock was one of the very things that we stored independent of the serialized world.

A new `GameSession` is created when the player starts a new game or loads a savegame.
Properties that you don't want to keep after loading a new world/savegame must go into `GameSession` I.e.: temporary effects introduced by cheats (i.e. gamespeed)

Have a look at:
https://github.com/REGoth-project/REGoth/blob/master/src/engine/GameSession.cpp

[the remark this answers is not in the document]
ataulien: Ah, I completely forgot about how useful the GameSession-class was. We absolutely need something similar very soon. However, in my opinion, we can merge the PR like this and move the clock to the session object in another PR.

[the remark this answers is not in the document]
markusobi: @ataulien I agree. I already wrote some documentation (not published yet) about how switching worlds can be modelled. In detail: when to load the world from memory (serialized), when to load it from disk and when to save it to memory (serialized) / disk. I will add this to our docs as soon as I am back from vacation.

ataulien:
@KirmesBude Can you rebase with the latest master so I can merge it (If you're okay with it)?

KirmesBude:
@ataulien I will rebase on master later today. 
I still want to do a thing or two for GameClock, but maybe that should be done in a seperate PR in the interest of not keeping this PR around for too long.

ataulien:
I agree, this branch already had some in-depth reviews :smile:

KirmesBude:
@ataulien You can go ahead and merge it.

PR #17, opened 2019-04-29
Serialization

ataulien:
This PR adds a Component to be attached to the scene root which manages the game world with (hopefully) full RTTI/Serialization foo. It also gets rid of the global `gGameScript()` and `gWorld()` objects by refactoring those into the `GameWorld`-component.

See is the documentation of the `GameWorld`-class for more information.


src/RTTI/RTTI_Waynet.hpp
```diff
@@ -0,0 +1,41 @@
...
+  class RTTI_Waynet : public bs::RTTIType<Waynet, bs::Component, RTTI_Waynet>
+  {
+    using UINT32 = bs::UINT32;
+
+    BS_BEGIN_RTTI_MEMBERS
+    BS_RTTI_MEMBER_REFL_ARRAY(mWaypoints, 0)
+    BS_END_RTTI_MEMBERS
+
+  public:
+    RTTI_Waynet()
```
KirmesBude: This all looks very similar in the RTTI files.
What keeps us from macro-fying this aswell?

ataulien: Not much I guess. Some RTTI-classes need to run code in the constructor, but most don't. 
I think the macros inside `RTTIUtils.hpp` help a lot in keeping the actual components code clean. I actually made myself some automated snippets for creating component headers, sources and their RTTI. 

We could also make a template for people to use to create new components.

KirmesBude: That's true for the constructor.
I was more talking about the other 3 methods.
Though if the point is readability, we would not gain much I guess.

docs-source/content/world.rst
```diff
@@ -143,8 +113,8 @@ As you can see, this is using an internal header, so handle it with care.
 Using the world
 ---------------
 
-Once you have successfully initialized a world, you can use ``REGoth::gWorld()``
-to access it. See ``src/world/GameWorld.hpp`` for the ``GameWorld``-class.
+Once you have successfully initialized a world, you can use created ``GameWorld``-component
```
mworchel: > you can use **the** created

Missing a word

docs/_sources/content/world.rst.txt
```diff
@@ -143,8 +113,8 @@ As you can see, this is using an internal header, so handle it with care.
 Using the world
 ---------------
 
-Once you have successfully initialized a world, you can use ``REGoth::gWorld()``
-to access it. See ``src/world/GameWorld.hpp`` for the ``GameWorld``-class.
+Once you have successfully initialized a world, you can use created ``GameWorld``-component
```
mworchel: Same missing word as above

src/RTTI/RTTIPlainTypes.hpp
```diff
@@ -0,0 +1,71 @@
...
+#include "RTTI_TypeIDs.hpp"
+
+namespace bs
+{
+  /**
+   * RTTIPlainType for std::pair.
+   *
+   * @see		RTTIPlainType
+   */
+  #if 0
```
mworchel: Since it is always "ifdeffed" out, do we need this code/this file?

src/RTTI/RTTIUtil.hpp
```diff
@@ -0,0 +1,144 @@
+#pragma once
+#include "RTTI_TypeIDs.hpp"
+#include <BsCorePrerequisites.h>
+#include <Private/RTTI/BsGameObjectRTTI.h>  // Says private, but bs:f uses this too in their RTTIs
+#include <Reflection/BsRTTIType.h>
+
+/**
+ * \file
```
mworchel: I was wondering, if these comment blocks describing the file (content) would be best placed at the top even above the includes. Since they are some kind of meta information, one usually wants to read these before encountering any code related stuff. What's your thought?

src/RTTI/RTTIUtil.hpp
```diff
@@ -0,0 +1,144 @@
...
+/**
+ * For use in the actual components header. Declares the functions for accessing
+ * the RTTI-instance of the component.
+ *
+ * Example:
+ *
+ *    class MyComponent : public bs::Component
+ *    {
+ *    public:
+ *      REGOTH_COMPONENT_DECLARE_RTTI(MyComponent)
```
mworchel: Macro is named `REGOTH_DECLARE_RTTI`

src/RTTI/RTTIUtil.hpp
```diff
@@ -0,0 +1,144 @@
...
+
+/**
+ * For use in the actual components source. Defines the functions for accessing
+ * the RTTI-instance of the component.
+ *
+ * Example:
+ *
+ *     namespace REGoth
+ *     {
+ *       REGOTH_COMPONENT_DEFINE_RTTI(MyComponent)
```
mworchel: Also here we don't need the `_COMPONENT`.

src/RTTI/RTTIUtil.hpp
```diff
@@ -0,0 +1,144 @@
...
+    return classname::getRTTIStatic();         \
+  }
+
+/**
+ * For use in the RTTI-class. Implements the glue code to create an instance of the component
+ * the RTTI describes.
+ *
+ * Use this if the RTTI describes a class derived from bs::Component.
+ */
+#define REGOTH_IMPLEMENT_RTTI_CLASS_FOR_COMPONENT(classname)  \
```
mworchel: Maybe mention the necessary includes for this macro so that classes like `bs::GameObjectRTTI` are found.

src/RTTI/RTTIPlainTypes.hpp
ataulien: This was me fighting with serializing complex `bs::Map` types. I will remove it.

src/RTTI/RTTIUtil.hpp
ataulien: I agree. Since this is a file which does not contain any class definitions, the comment should be at the top.

src/RTTI/RTTIPlainTypes.hpp
mworchel: Doesn't it work out of the box, somehow?

src/components/GameWorld.hpp
```diff
@@ -0,0 +1,301 @@
...
+   *
+   * Example to save a World:
+   *
+   *    gameWorld->save("MySavegame");
+   *
+   *
+   * Example to load a saved World:
+   *
+   *    bs::HPrefab prefab = GameWorld::load("MySavegame");
+   *    prefab->instantiate();
```
mworchel: The documentation says that 

> [...] This is done by calling Prefab::instantiate() which returns a SceneObject. By default this scene object will be parented to the current scene root, but can then be manipulated as any other scene object.

What happens if I save the current scene root and perform a load operation directly after? Will the previously saved root be parented to the current scene root (instead of replacing it), resulting in "two" GameWorld components being active at the same time?

src/RTTI/RTTIPlainTypes.hpp
ataulien: It works for plain types, such as `bs::Map<int, int>` or even `bs::Map<bs::String, int>`. It even does containers like `bs::map<bs::String, bs::Vector<int>>`. But you're out of luck if you're mapping a generic `IReflectable` such as `ScriptObject`.

src/components/GameWorld.hpp
ataulien: I think there are special *Scene*-prefabs which will replace the root. But not that I think of it, I remember seeing a commit which changed the way to load scenes from using a prefab to going through the `SceneManager` or what's it called. I will look it up later.

src/components/NeedsGameWorld.hpp
```diff
@@ -0,0 +1,49 @@
...
+
+namespace REGoth
+{
+  class GameWorld;
+  using HGameWorld = bs::GameObjectHandle<GameWorld>;
+
+  /**
+   * Base-class for all components needing a GameWorld to work.
+   */
+  class NeedsGameWorld : public bs::Component
```
mworchel: Just stumbeled over the unusual class name. Maybe just personal taste. What about `GameWorldDependent`?

src/components/NeedsGameWorld.cpp
```diff
@@ -0,0 +1,41 @@
...
+#include <RTTI/RTTI_NeedsGameWorld.hpp>
+#include <components/GameWorld.hpp>
+
+namespace REGoth
+{
+  NeedsGameWorld::NeedsGameWorld(const bs::HSceneObject& parent)
+      : bs::Component(parent)
+  {
+    setName("NeedsGameWorld");
+    findGameWorld();
```
mworchel: Maybe move the logic to `onInitialized()` (although it shouldn't make a big difference)

src/RTTI/RTTIUtil.hpp
```diff
@@ -0,0 +1,144 @@
...
+                                                              \
+  bs::SPtr<bs::IReflectable> newRTTIObject() override         \
+  {                                                           \
+    return bs::GameObjectRTTI::createGameObject<classname>(); \
```
mworchel: Could this be the reason why making the default constructors protected does not work, as the static method of GameObjectRTTI tries to access the protected constructor and not the RTTI class itself?

mworchel: In the newer bs::f version they seem to use `SceneObject::createEmptyComponent()` that's why they declare SceneObject as friend in the component classes.

src/components/NeedsGameWorld.hpp
ataulien: It is unusual, but I feel it explains pretty nicely what you get if you use the component. There is the issue that a name like `GameWorldDependent` sounds like the component itself *is* a `GameWorld` which was extended with some other functionality. There could be a `GameWorldWithScriptInit` or `GameWorldIndoor` into whose naming scheme the component would incorrectly fall.

I agree classes should be named after a *thing*, but I'd have to research more about the naming of *components*, as they're not really a class in the traditional OOP way but more like traits assigned to an object and I would absolutely call a trait `NeedsGameWorld`.

src/RTTI/RTTIUtil.hpp
ataulien: Accessing private and protected members from static functions should work fine, though. I haven't used friend classes much however...

mworchel: Nope, protected and private members can only be accessed if you explicitely grant access to them. Static functions are not exception to this.

ataulien: How do you explicitly grant access to them? Is that only a thing for friend-classes, because it works fine from within classes: https://godbolt.org/z/f5_2Z9

Ah, and now I understood what you meant! `createGameObject` is a template accessing the constructor, so the compiler *could* decide that `GameObjecrRTTI` is not a friend and has no access to it. Took me a minute, sorry :smile:

src/components/NeedsGameWorld.hpp
mworchel: If you see them as trait, I can follow your arguments. I'm totally fine with the decision but was curious for the reason as I'm so used usual class naming conventions. The closest I would have guessed would be `NeedingGameWorld` or similar because `NeedsGameWorld` sounds more like the name of a (boolean) variable.

src/RTTI/RTTIUtil.hpp
mworchel: Yes, exactly! And in the current master it seems like they moved the static function for creating empty components to the `SceneObject` class and therefore they made `SceneObject` friend to all components.

src/components/NeedsGameWorld.hpp
ataulien: We can follow "proper" OOP practices and call it `RequireGameWorldManagerInterfaceFactory`. :smile:  While working with Unity for university, my fellow students made sure to get a  `Manager` into each components name!

Jokes aside, I'll research some component naming when I get home!

mworchel: You surely meant the `RequireGameWorldManagerInterfaceFactoryVisitorObserver`. Because you want to make sure not to miss any design patterns and use them all in each circumstance 😀 I vote for this name

docs/_sources/content/world.rst.txt
ataulien: That is actually build output, so this is automatically resolved when fixing the other one.

src/RTTI/RTTIUtil.hpp
ataulien: Done.

src/RTTI/RTTIUtil.hpp
ataulien: Done

src/RTTI/RTTIUtil.hpp
ataulien: Done.

src/RTTI/RTTIUtil.hpp
ataulien: I'll update bsf to fix this for good.

src/RTTI/RTTIPlainTypes.hpp
ataulien: Removed.

src/components/NeedsGameWorld.hpp
ataulien: Can't find proper guides on component naming. Only thing I have read was "Should be named after what it does", like `MeshCollider`. Now, what does that component do? I don't know anymore. If you're ok with it I'd like to keep the name. I kinda like it now :smile:

src/components/GameWorld.hpp
ataulien: There is `gSceneManager().save(...)` but that also saves stuff like the FPS-Camera, which does not have RTTI. We'll figure that out in another PR, if that's okay for you.

src/main_WorldCacheTest.cpp
```diff
@@ -0,0 +1,64 @@
...
+  }
+
+  void setupScene() override
+  {
+    using namespace REGoth;
+
+    const bs::String world = "OLDWORLD.ZEN";
+    const bs::String saveGame = "MySafeGame";
+
+    bs::Timer timer;
```
KirmesBude: I assume you had this in here to see how much time it takes?
I think it is fine to actually use it her, since it is a TestApp.

src/components/CharacterKeyboardInput.hpp
```diff
@@ -34,16 +35,10 @@ namespace REGoth
...
     HCharacterAI mCharacterAI;
 
-    /************************************************************************/
-    /* RTTI                                                                 */
-    /************************************************************************/
   public:
-    friend class RTTI_CharacterKeyboardInput;
-    static bs::RTTITypeBase* getRTTIStatic();
-    bs::RTTITypeBase* getRTTI() const override;
+    REGOTH_DECLARE_RTTI(Character);
```
KirmesBude: Missing the new macros in CharacterKeyboardInput.cpp

src/components/VisualCharacter.hpp
```diff
@@ -76,17 +74,11 @@ namespace REGoth
...
     BodyState mBodyState;
 
-    /************************************************************************/
-    /* 								RTTI */
-    /************************************************************************/
   public:
-    friend RTTI_VisualCharacter;
-    static bs::RTTITypeBase* getRTTIStatic();
-    bs::RTTITypeBase* getRTTI() const override;
+    REGOTH_DECLARE_RTTI(Character);
```
KirmesBude: Missing the new macros in VisualCharacter.cpp

src/components/VisualInteractiveObject.hpp
```diff
@@ -21,19 +21,11 @@ namespace REGoth
...
-  private:
-
-    /************************************************************************/
-    /* RTTI                                                                 */
-    /************************************************************************/
   public:
-    friend class RTTI_VisualInteractiveObject ;
-    static bs::RTTITypeBase* getRTTIStatic();
-    bs::RTTITypeBase* getRTTI() const override;
+    REGOTH_DECLARE_RTTI(Character);
```
KirmesBude: Missing the new macros in VisualInteractiveObject.cpp

mworchel:
Compiles and the world is successfully saved/loaded. However, in the world viewer I cannot move after loading and the character's head is attached to its stomach. And also the problem with the physics mesh persists (character falling indefinitely). Any idea what the heck this is causing? Is it a bs::f issue?

ataulien:
I'll merge this and open issues for the remaining problems.

PR #21, opened 2019-05-05
Bugfixes 1

ataulien:
Collection of small bugfixes. Look at the commit messages for more information.

### DIFF
--- a/src/components/VisualSkeletalAnimation.cpp
+++ b/src/components/VisualSkeletalAnimation.cpp
@@ -53,10 +53,14 @@ namespace REGoth
     deleteObjectSubtree();
     mModelScript = modelScript;
 
+    createAnimationMap();
+  }
+
+  void VisualSkeletalAnimation::createAnimationMap()
+  {
     for (bs::HAnimationClip clip : mModelScript->getAnimationClips())
     {
       mAnimationClips[clip->getName()] = clip;
-      // bs::gDebug().logDebug(clip->getName());
     }
   }
 
@@ -298,7 +302,13 @@ namespace REGoth
   {
     bs::HAnimationClip clip;
 
-    if (Animation::isTransitionNeeded(state))
+    throwIfNotReadyForRendering();
+
+    if (!mSubAnimation->isPlaying())
+    {
+      clip = findAnimationClip(state);
+    }
+    else if (Animation::isTransitionNeeded(state))
     {
       bs::String from = getStateFromPlayingAnimation();
       bs::String to   = Animation::getStateName(state);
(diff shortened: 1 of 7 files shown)


src/components/VisualSkeletalAnimation.cpp
```diff
@@ -53,10 +53,14 @@ namespace REGoth
     deleteObjectSubtree();
     mModelScript = modelScript;
 
+    createAnimationMap();
+  }
+
+  void VisualSkeletalAnimation::createAnimationMap()
+  {
     for (bs::HAnimationClip clip : mModelScript->getAnimationClips())
```
mworchel: Shouldn't we make sure that `mModelScript` is valid in this function as it could be called from `onDeserializationEnded`?

src/components/VisualSkeletalAnimation.cpp
```diff
@@ -298,7 +300,13 @@ namespace REGoth
   {
     bs::HAnimationClip clip;
 
-    if (Animation::isTransitionNeeded(state))
+    throwIfNotReadyForRendering();
+
+    if (!mSubAnimation->isPlaying())
+    {
+      clip = findAnimationClip(state);
```
mworchel: Haven't digged into how animations work but isn't that `if`-case already handled with the `else`? It's the same code

src/components/VisualSkeletalAnimation.cpp
ataulien: Good point. I'll put a safety-check.

src/components/VisualSkeletalAnimation.cpp
ataulien: Not quite. If there is no animation playing, then `getStateFromPlayingAnimation()` will return an empty string on the next branch, thus, no transition animation can be found. I didn't want to add conditions to the next `if` to make it clear what is supposed to happen. I'll refactor it!

ataulien:
@mworchel How's this?

mworchel:
It's much better, especially with the comments.

However, I tried out the branch and it seems like everything is fine if you run the WorldViewer for the first time but the second time (when the "savegame" is loaded) there seems to be a problem with the animations. If you move forward once it again plays the `T_RUN_2_RUNL` animation and cannot transition away from it anymore.

ataulien:
How up-to-date is your bsf? I'll try that, but I thought it worked fine for me...

mworchel:
Using the bsf master.

ataulien:
Okay, can reproduce it. What the F?

ataulien:
@mworchel Should work now.

mworchel:
Works 👍

PR #22, opened 2019-05-09
Show armor

ataulien:
Changes:
 - Fixes Skeletal-Meshes broken when the source files supply their own skeleton
 - Disables physics on all characters not controlled by the player for performance reasons,
 - Dresses the Characters with the Armor equipped inside their instance constructor

### DIFF
--- a/src/components/CharacterAI.cpp
+++ b/src/components/CharacterAI.cpp
@@ -38,6 +38,16 @@ namespace REGoth
     }
   }
 
+  void CharacterAI::deactivatePhysics()
+  {
+    mIsPhysicsActive = false;
+  }
+
+  void CharacterAI::activatePhysics()
+  {
+    mIsPhysicsActive = true;
+  }
+
   bool CharacterAI::goForward()
   {
     if (!isStateSwitchAllowed()) return false;
@@ -76,6 +86,7 @@ namespace REGoth
 
     mTurnDirection = TurnDirection::Left;
 
+    // return mVisual->tryPlayTransitionAnimationTo("T_RUNTURNL");
     return true;
   }
 
@@ -85,6 +96,7 @@ namespace REGoth
 
     mTurnDirection = TurnDirection::Right;
 
+    // return mVisual->tryPlayTransitionAnimationTo("T_RUNTURNR");
     return true;
   }
 
@@ -97,6 +109,11 @@ namespace REGoth
 
   bool CharacterAI::stopTurning()
   {
+    // if (mTurnDirection != TurnDirection::None)
+    // {
+    //   mVisual->playAnimation(mVisual->findAnimationClip("S_RUN"));
+    // }
+
     mTurnDirection = TurnDirection::None;
 
     return true;
@@ -127,6 +144,12 @@ namespace REGoth
 
   void CharacterAI::fixedUpdate()
   {
+    if (!mIsPhysicsActive)
+    {
+      // Skip all movement and physics calculations to save processing time.
+      return;
+    }
+
     if (isTurningAllowed())
     {
       handleTurning();
--- a/src/components/CharacterAI.cpp
+++ b/src/components/CharacterAI.cpp
@@ -86,7 +86,6 @@ namespace REGoth
 
     mTurnDirection = TurnDirection::Left;
 
-    // return mVisual->tryPlayTransitionAnimationTo("T_RUNTURNL");
     return true;
   }
 
@@ -95,8 +94,6 @@ namespace REGoth
     if (!isTurningAllowed()) return false;
 
     mTurnDirection = TurnDirection::Right;
-
-    // return mVisual->tryPlayTransitionAnimationTo("T_RUNTURNR");
     return true;
   }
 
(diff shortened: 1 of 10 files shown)


src/scripting/daedalus/DaedalusVMForGameWorld.cpp
```diff
@@ -292,6 +292,24 @@ namespace REGoth
       bs::INT32 bodyTexIndex = popIntValue();
       bs::String bodyMesh    = popStringValue();
 
+      // If an armor is set here, we need to replace the body mesh from the input parameters with the
+      // visual settings from inside the armor instance
+      if (armorInstance != -1)
+      {
+        // TODO: Original Gothic adds the Armor straight to the inventory. We just create a
+        // temporary instance to extract the visual information from it
+        ScriptObjectHandle armorSObj = getInstance(armorInstance);
```
mworchel: `armorInstance` is a `bs::UINT32` but `getInstance` expects a `bs::String`, so it doesn't compile for me due to this line

src/main_WorldViewer.cpp
```diff
@@ -32,11 +32,11 @@ class REGothWorldViewer : public REGoth::REGothEngine
     {
       world = GameWorld::importZEN(WORLD);
 
-      HCharacter hero = world->insertCharacter("PC_HERO", "STARTPOINT");
+      HCharacter hero = world->insertCharacter("PC_HERO", "START_DRAGONISLAND");
```
mworchel: `"START_DRAGONISLAND"` only works for Dragon Island I suppose? :D

src/components/CharacterAI.cpp
```diff
@@ -76,6 +86,7 @@ namespace REGoth
 
     mTurnDirection = TurnDirection::Left;
 
+    // return mVisual->tryPlayTransitionAnimationTo("T_RUNTURNL");
```
mworchel: Did you try to run the turning animations but it doesn't work? Is this code supposed to stay commented out or even stay at all?

src/components/CharacterAI.cpp
```diff
@@ -85,6 +96,7 @@ namespace REGoth
 
     mTurnDirection = TurnDirection::Right;
 
+    // return mVisual->tryPlayTransitionAnimationTo("T_RUNTURNR");
```
mworchel: Same as line 89

src/components/CharacterAI.cpp
```diff
@@ -97,6 +109,11 @@ namespace REGoth
 
   bool CharacterAI::stopTurning()
   {
+    // if (mTurnDirection != TurnDirection::None)
+    // {
+    //   mVisual->playAnimation(mVisual->findAnimationClip("S_RUN"));
```
mworchel: Same as line 89

src/components/CharacterAI.cpp
ataulien: The code is correct, but there is a bug with the animation importer which makes turning animations look weird. I can remove those lines from the PR. 
I agree that commented out code without any note as to *why* it was commented out isn't great.

src/scripting/daedalus/DaedalusVMForGameWorld.cpp
ataulien: Weird, did I forget to commit something?

src/main_WorldViewer.cpp
ataulien: A method to get the world starting point from the `GameWorld` would be nice to have I guess.

src/scripting/daedalus/DaedalusVMForGameWorld.cpp
ataulien: There seems to be more missing code, actually. Now, I wonder where that went...

ataulien: Can't find it in stashes, it's not on other branches, guess I'lll write that again.

ataulien: Actually I found those commits! Seems like I forgot to push them before merging https://github.com/REGoth-project/REGoth-bs/pull/21...

src/scripting/daedalus/DaedalusVMForGameWorld.hpp
```diff
@@ -93,7 +99,7 @@ namespace REGoth
     public:
       REGOTH_DECLARE_RTTI(DaedalusVMForGameWorld);
 
-    public:                              // FIXME: RTTI protected
+    public:                                // FIXME: RTTI protected
```
ataulien: We should take care of the remaining RTTI-protected FIXME in another PR. It's not straightforward for bs::IReflectable. We also befriend bs::SceneObject inside REGOTH_DECLARE_RTTI, which is only correct for components.

mworchel: We probably need separate `DECLARE` and `DEFINE` macros for `IReflectable` and  `Component` (and probably also a seperate one for pure abstract `IReflectables`). Doing it for `IReflectable` is pretty straight-forward (see `RaycastShadowSampler` in the shadow-sampling branch) and could also be encapsulated in a macro.

src/main_WorldViewer.cpp
mworchel: Is this something one could easily get from the waynet? Like does it always contain the string `"START"` or something?

mworchel:
Compiles and works 👍 
What about the start point stuff?

ataulien:
That's something for an other PR. Right now the player is just dropped at position (0,0,0) if the startpoint can't be found. I guess that would be another good-first-issue :smile:

PR #24, opened 2019-05-10
File system changes, Mod-files and Camera

ataulien:
This PR adds some more functionality to the File-System utilities and allows the user to load Mod-files via overriding a virtual method in `REGothEngine`.

It also adds a simple third-person camera.


src/original-content/OriginalGameFiles.cpp
```diff
@@ -60,6 +65,11 @@ namespace REGoth
     return filterFilesInDirectoryByExt(dataDirectory(), ".vdf");
   }
 
+  bs::Vector<bs::Path> OriginalGameFiles::allModPackages() const
+  {
+    return filterFilesInDirectoryByExtRecursive(modDirectory(), ".mod");
```
mworchel: `*.mod` or `*.modvdf`? The documentation of `loadGamePackages` says `*.modvdf`

src/original-content/VirtualFileSystem.cpp
```diff
@@ -89,11 +89,37 @@ bs::Vector<bs::String> VirtualFileSystem::listAllFiles()
...
+  enum
+  {
+    RespectCase = false,
+    LowerCase  = true,
+  };
+
+  bs::Vector<bs::String> result;
+  for(const auto& f : all)
+  {
+    if (bs::StringUtil::endsWith(f, ext, RespectCase))
```
mworchel: Why is the case respected in this case? I think it's something which can cause a lot of confusion and is very unintuitive for people who don't carefully read the documentation of this method.

src/components/ThirdPersonCamera.hpp
```diff
@@ -0,0 +1,99 @@
...
+     * Applies the given delta to the cameras distance. This can be used to zoom in the
+     * camera using the mousewheel, for example.
+     *
+     * There is no direct setter for this distance, as the distance has to adapt to
+     * certain situations during gameplay.
+     *
+     * @param  delta  Positive values will move the camera further away from the target,
+     *                Negative values will move it closer.
+     */
+    void modifyDistance(float delta);
```
mworchel: Misses the implementation in the `*.cpp`. Maybe add an empty body there with a `TODO` as we will certainly need this.

mworchel: You probably also had in mind the case where the camera is moved towards the character when back-facing a wall or any other object that is closer than the camera distance. Maybe we could use bs::f's physics to solve this by giving the camera SO a small collider and constraining the motion of this collider to the axis between camera and player. Think of a spring.

mworchel:
Only some minor nitpicks, otherwise compiles and runs without problems. The third person camera turned out pretty good! 

One more thing:
Although not related to this PR, could you remove the RTTI includes in `RTTI_GameGlock.hpp`
```cpp
#include "RTTI_TypeIDs.hpp"
#include <BsCorePrerequisites.h>
#include <Private/RTTI/BsGameObjectRTTI.h>  // Says private, but bs:f uses this too in their RTTIs
#include <Reflection/BsRTTIType.h>
```
 and replace them with our `RTTIUtil.hpp`

src/original-content/OriginalGameFiles.cpp
ataulien: `.mod` is correct. Good catch.

src/original-content/VirtualFileSystem.cpp
ataulien: All file names searched there are already in uppercase, as all methods from the virtual file system work with those. I chose to not use `ConvertToLower` here for efficiency, because it would then have to convert the filenames to lowercase as well. However, converting the extension to uppercase before the loop would work I guess.

src/components/ThirdPersonCamera.hpp
ataulien: Right, seems like I just forgot to implement that. In the original engine, the camera can be blocked by walls too, but they do it via a raycast as far as I know.

ataulien:
> Although not related to this PR, could you remove the RTTI includes in RTTI_GameGlock.hpp

Will do that as well, while I'm at it :smile:

src/original-content/VirtualFileSystem.cpp
mworchel: Ah, I probably looked to quickly and didn't notice you are using the `endsWith` method (thought you were just plainly comparing strings and `f` is the extension). Converting the extension should work best, yes (I wouldn't worry so much about efficieny yet)

src/components/ThirdPersonCamera.hpp
ataulien: By the way, that method was more for zooming via the scrollwheel.

mworchel:
No more complaints. It is ready to be merged (after #22).

PR #25, opened 2019-05-12
Fix typo in RTTI declaration

mworchel:
Just fixing a small typo in the RTTI declaration of `VisualInteractiveObject` I discovered when toying around with the RTTI macros. Funny enough, it worked before.

### DIFF
--- a/src/RTTI/RTTIUtil.hpp
+++ b/src/RTTI/RTTIUtil.hpp
@@ -37,11 +37,11 @@
  * note that there is not semicolon after the makro!
  * Futhermore, the makro invocation needs to be in `public`-land of the component.
  */
-#define REGOTH_DECLARE_RTTI(classname)      \
-  friend class bs::SceneObject;             \
-  friend class RTTI_##classname;            \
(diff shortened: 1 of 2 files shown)


ataulien:
Can we make this being detected by the compiler somehow?

mworchel:
It was in fact discovered by the compiler as I changed the `REGOTH_DECLARE_RTTI` macro. The only possible way that comes into my mind (and which triggered it here for me) is accessing a member/method of the class passed to the macro. The compiler then bails out when you use the wrong class (whos header is *hopefully* not included). Something simple like
```cpp
using _TypeCheck = decltype(&classname::getRTTI);  
```
at the end of the macro works, however it is equally ugly.

ataulien:
Can you add something like this to the PR?

mworchel:
I added something like this with the latest commit and it's not even that ugly. It basically makes sure, that `getRTTI` has the same return type as `getRTTIStatic` while checking if the passed class name is known in the current file.

ataulien:
Good job! That's indeed a very nice solution.

PR #29, opened 2019-05-14
Vob Collision

ataulien:
This PR adds collision meshes to Objects which have the "Collide with Stuff"-flag set in the ZEN-file (`cdDyn`).

Might want to move this to a different place later, so we can dynamically instantiate objects with collision more easily, so this is kept rather simple.

### DIFF
--- a/src/components/GameWorld.cpp
+++ b/src/components/GameWorld.cpp
@@ -46,7 +46,12 @@ namespace REGoth
     if (!mZenFile.empty())
     {
       // Import the ZEN and add all scene objects as children to this SO.
-      Internals::constructFromZEN(SO(), mZenFile);
+      bs::HSceneObject so = Internals::constructFromZEN(SO(), mZenFile);
+
+      if (!so)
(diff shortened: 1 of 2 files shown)


mworchel:
Code looks fine to me and collision seems to work at least in most cases. However I encountered an object with which the hero didn't collide but I would suspect it to be collidable (see image attached). It's in `ADDONWORLD.ZEN` just left when you leave the temple where you start from.

collision

ataulien:
This could either be a bug on our side or Gothic's devs forgot to set the collision-flag for that object... Though, I don't remember clipping through objects too often in the original game.

mworchel:
We could file this as bug so it is not forgotten. One probably would have to check that by firing up the original game.

ataulien:
Sounds good!

(I can reproduce it by the way)

ataulien:
Issue is here: https://github.com/REGoth-project/REGoth-bs/issues/30

PR #31, opened 2019-05-15
Character AI foundation

ataulien:
This PR implements most of the boilerplate code needed to get the AI working. This includes:

 * Character Event Queue
 * Simple Pathfinding
 * Daily routines

Resource caching should also be improved now so seeing black textures should happen less often.

To test this, I have modified the World-Viewer sample to load `OLDMINE.ZEN` from Gothic I. There, you should see a bunch of Characters running to their daily routine spots. They may go off their paths and fall down though, which is a problem with the path optimization. Should be less of a problem on flat terrain.

### DIFF
--- a/src/AI/EventMessage.hpp
+++ b/src/AI/EventMessage.hpp
@@ -1,7 +1,9 @@
 #pragma once
 
+#include <RTTI/RTTIUtil.hpp>
 #include <BsCorePrerequisites.h>
 #include <Math/BsVector3.h>
+#include <scripting/ScriptTypes.hpp>
 
 namespace REGoth
 {
@@ -42,7 +44,7 @@ namespace REGoth
     /**
      * Basic event-message. Contains type.
      */
-    struct EventMessage// : bs::IReflectable // TODO: Implement RTTI
+    struct EventMessage : public bs::IReflectable
     {
       using SharedEMessage = std::shared_ptr<EventMessage>;
 
@@ -115,6 +117,8 @@ namespace REGoth
        * External callbacks to trigger if this message gets processed.
        */
       bs::Event<void(SharedEMessage)> onMessageDone;
+
+      REGOTH_DECLARE_RTTI_FOR_REFLECTABLE(EventMessage)
     };
 
     struct NpcMessage : public EventMessage
@@ -128,6 +132,8 @@ namespace REGoth
       virtual ~NpcMessage()
       {
       }
+
+      REGOTH_DECLARE_RTTI_FOR_REFLECTABLE(NpcMessage)
     };
 
     struct DamageMessage : public NpcMessage
@@ -146,6 +152,8 @@ namespace REGoth
       }
 
       // TODO: Implement DamageDescriptor!
+
+      REGOTH_DECLARE_RTTI_FOR_REFLECTABLE(DamageMessage)
     };
 
     struct WeaponMessage : public NpcMessage
@@ -181,12 +189,14 @@ namespace REGoth
       // EWeaponMode targetMode;
 
       // TODO: There is an animation-field here, find out what it does!
+
+      REGOTH_DECLARE_RTTI_FOR_REFLECTABLE(WeaponMessage)
     };
 
     struct MovementMessage : public NpcMessage
     {
       enum MovementSubType
-      { 
+      {
         ST_RobustTrace,
         ST_GotoPos,
         ST_GotoObject,
@@ -248,6 +258,8 @@ namespace REGoth
        * General purpose mode
        */
       int targetMode = 0;
+
+      REGOTH_DECLARE_RTTI_FOR_REFLECTABLE(MovementMessage)
     };
 
     struct AttackMessage : public NpcMessage
@@ -273,13 +285,13 @@ namespace REGoth
       {
         messageType    = EventMessageType::Attack;
         startFrame     = 0.0f;
-        animationIndex = static_cast<size_t>(-1);
+        animationIndex = -1;
       }
 
       /**
        * How far we are in our combo
        */
-      int combo;
+      bs::INT32 combo;
 
       /**
        * Object we are attacking
@@ -289,12 +301,14 @@ namespace REGoth
       /**
        * Animation to use when attacking, -1 if none.
        */
-      size_t animationIndex;
+      bs::INT32 animationIndex;
 
       /**
        * Frame to skip the animation to
        */
       float startFrame;
+
+      REGOTH_DECLARE_RTTI_FOR_REFLECTABLE(AttackMessage)
     };
 
     struct UseItemMessage : public NpcMessage
@@ -315,6 +329,8 @@ namespace REGoth
       // TODO: We have no actual object for these right now, as they are explicitly in the
       // inventory.
       // TODO: Use the itemhandle here or something
+
+      REGOTH_DECLARE_RTTI_FOR_REFLECTABLE(UseItemMessage)
     };
 
     struct StateMessage : public NpcMessage
@@ -339,7 +355,7 @@ namespace REGoth
       /**
        * Symbol to the Setupfunction for this state (ZS_...)
        */
-      size_t functionSymbol;
+      Scripting::SymbolIndex functionSymbol;
 
       /**
        * Whether the old state should be ended properly (true), or just interrupted (false)
@@ -371,6 +387,8 @@ namespace REGoth
        * Waypoint name to got to, in case the state needs that
        */
       bs::String wpname;
+
+      REGOTH_DECLARE_RTTI_FOR_REFLECTABLE(StateMessage)
     };
 
     class ManipulateMessage : public NpcMessage
@@ -438,12 +456,14 @@ namespace REGoth
       /**
        * State to go to when using the item
        */
-      int targetState;
+      bs::INT32 targetState;
 
       /**
        * Animation to play, for example on pickup
        */
       bs::String animation;
+
+      REGOTH_DECLARE_RTTI_FOR_REFLECTABLE(ManipulateMessage)
     };
 
     struct ConversationMessage : public NpcMessage
@@ -493,7 +513,6 @@ namespace REGoth
         messageType    = EventMessageType::Conversation;
         status         = Status::INIT;
         canceled       = false;
-        waitIdentifier = nullptr;
       }
 
       /**
@@ -531,18 +550,12 @@ namespace REGoth
       /**
        * Animation to play. -1 if none
        */
-      size_t animationIndex;
+      bs::INT32 animationIndex;
 
       /**
        * Animation name to be used. If empty, index will be checked.
        */
       bs::String animation;
-
-      /**
-       * SharedPointer to the message we are waiting for. Currently not used. Useful for debug
-       * purpose
-       */
-      SharedEMessage waitIdentifier;
 
       /**
        * Stores the state of the message, to handle multiple stages/states inside the
@@ -560,6 +573,8 @@ namespace REGoth
        */
       // FIXME: Reimplement that using a sound handle
       // Utils::Ticket<World::AudioWorld> soundTicket;
+
+      REGOTH_DECLARE_RTTI_FOR_REFLECTABLE(ConversationMessage)
     };
 
     struct MagicMessage : public NpcMessage
@@ -586,6 +601,8 @@ namespace REGoth
       }
 
       // TODO: Implement
+
+      REGOTH_DECLARE_RTTI_FOR_REFLECTABLE(MagicMessage)
     };
 
     struct MobMessage : public EventMessage
@@ -624,6 +641,8 @@ namespace REGoth
        * Whether to play an animation.
        */
       bool playAnimation;
+
+      REGOTH_DECLARE_RTTI_FOR_REFLECTABLE(MobMessage)
     };
-  }  // namespace EventMessages
-}  // namespace Logic
+  }  // namespace AI
+}  // namespace REGoth
--- a/src/AI/EventMessage.hpp
+++ b/src/AI/EventMessage.hpp
@@ -8,19 +8,19 @@
 
 namespace REGoth
 {
+  class Waypoint;
+  using HWaypoint = bs::GameObjectHandle<Waypoint>;
+
+  class Character;
+  using HCharacter = bs::GameObjectHandle<Character>;
+
+  class Item;
+  using HItem = bs::GameObjectHandle<Item>;
+
   namespace AI
   {
     struct EventMessage;
     using SharedEMessage = bs::SPtr<EventMessage>;
-
-    class Waypoint;
-    using HWaypoint = bs::GameObjectHandle<Waypoint>;
-
-    class Character;
-    using HCharacter = bs::GameObjectHandle<Character>;
-
-    class Item;
-    using HItem = bs::GameObjectHandle<Item>;
 
     enum class EventMessageType
     {
@@ -344,14 +344,14 @@ namespace REGoth
       }
 
       /**
-       * Symbol to the Setupfunction for this state (ZS_...)
+       * Name of the Setupfunction for this state (ZS_...)
        */
-      Scripting::SymbolIndex functionSymbol;
+      bs::String state;
 
       /**
-       * Whether the old state should be ended properly (true), or just interrupted (false)
+       * Whether the old state should be ended properly, or be interrupted and canceled.
        */
-      bool endOldState;
+      bool interruptOldState = false;
 
       /**
        * Symbols for other and victim
@@ -362,17 +362,17 @@ namespace REGoth
       /**
        * Whether this belongs to the daily routine
        */
-      bool isRoutineState;
+      bool isRoutineState = false;
 
       /**
        * Whether this is a program-managed state. List of states in NpcScriptState.h
        */
-      bool isPrgState;
+      bool isPrgState = false;
 
       /**
        * If this is a wait-message, this is how long we should wait
        */
-      float waitTime;
+      float waitTime = 0.0f;
 
       /**
        * Waypoint name to got to, in case the state needs that
@@ -539,9 +539,9 @@ namespace REGoth
       bs::Vector3 targetPosition;
 
       /**
-       * Animation to play. -1 if none
+       * Only valid while the message is being processed. Clip matching the given animation name.
        */
-      bs::INT32 animationIndex;
+      bs::HAnimationClip playingClip;
 
       /**
        * Animation name to be used. If empty, index will be checked.
--- a/src/AI/EventMessage.hpp
+++ b/src/AI/EventMessage.hpp
@@ -128,12 +128,6 @@ namespace REGoth
       virtual ~NpcMessage()
       {
       }
-
-      /**
-       * Name of the object this message targets. This could be the attack target or a waypoint
-       * to go to.
-       */
-      bs::String targetObjectName;
     };
 
     struct DamageMessage : public NpcMessage
@@ -227,8 +221,6 @@ namespace REGoth
       MovementMessage()
       {
         messageType = EventMessageType::Movement;
-        walkMode    = WalkMode::Run;
-        targetMode  = 0;
       }
 
       /**
@@ -250,12 +242,12 @@ namespace REGoth
       /**
        * How we should go to our position
        */
-      WalkMode walkMode;
+      WalkMode walkMode = WalkMode::Run;
 
       /**
        * General purpose mode
        */
-      int targetMode;
+      int targetMode = 0;
     };
 
     struct AttackMessage : public NpcMessage
--- a/src/AI/EventMessage.hpp
+++ b/src/AI/EventMessage.hpp
@@ -1,8 +1,9 @@
 #pragma once
 
-#include <RTTI/RTTIUtil.hpp>
 #include <BsCorePrerequisites.h>
+#include <AI/WalkMode.hpp>
 #include <Math/BsVector3.h>
+#include <RTTI/RTTIUtil.hpp>
 #include <scripting/ScriptTypes.hpp>
 
 namespace REGoth
@@ -218,16 +219,6 @@ namespace REGoth
         ST_MoveMax
       };
 
-      enum class WalkMode
-      {
-        Run = 0,
-        Walk,
-        Sneak,
-        Water,
-        Swim,
-        Dive
-      };
-
       MovementMessage()
       {
         messageType = EventMessageType::Movement;
@@ -252,7 +243,7 @@ namespace REGoth
       /**
        * How we should go to our position
        */
-      WalkMode walkMode = WalkMode::Run;
+      AI::WalkMode walkMode = AI::WalkMode::Run;
 
       /**
        * General purpose mode
@@ -335,14 +326,14 @@ namespace REGoth
 
     struct StateMessage : public NpcMessage
     {
-      enum TStateSubType : bs::UINT32
+      enum StateSubType : bs::UINT32
       {
-        EV_StartState,
-        EV_Wait,
-        EV_SetNpcsToState,
-        EV_SetTime,
-        EV_ApplyTimedOverlay,
-        EV_StateMax
+        ST_StartState,
+        ST_Wait,
+        ST_SetNpcsToState,
+        ST_SetTime,
+        ST_ApplyTimedOverlay,
+        ST_StateMax
       } ConversationMessage;
 
       StateMessage()
@@ -510,9 +501,9 @@ namespace REGoth
 
       ConversationMessage()
       {
-        messageType    = EventMessageType::Conversation;
-        status         = Status::INIT;
-        canceled       = false;
+        messageType = EventMessageType::Conversation;
+        status      = Status::INIT;
+        canceled    = false;
       }
 
       /**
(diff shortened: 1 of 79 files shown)


src/AI/EventMessage.hpp
```diff
@@ -0,0 +1,648 @@
...
+      {
+      }
+
+      REGOTH_DECLARE_RTTI_FOR_REFLECTABLE(NpcMessage)
+    };
+
+    struct DamageMessage : public NpcMessage
+    // FIXME: Reimplement that using a sound handle
+    {
+      enum class DamageSubType
```
mworchel: May want to follow the naming conventions here aswell like `ST_Once` etc. and also make it a plain `enum`, not `enum class`, otherwise casting to `unsigned int` will be a bit cumbersome. One could also explicitely derive the enum from `unsigned int` for it to be even more clear
```cpp
enum DamageSubType : unsigned int
{
  ST_Once,
  ST_PerFrame,
  ST_Max
};
```
Would have to do it for the others aswell.

ataulien:
Before I forget: You'd need to update bsf to latest master for some rtti related things. I just didn't find the time to commit the update and address the review comments yet

mworchel:
Thanks for the hint with bs::f. Say, is this supposed to successfully run all worlds with init scripts, at least without crashing? When loading `ADDONWORLD.ZEN` I get an exception

> [ERROR] A fatal error occurred and the program has to terminate!
>   - Error: InvalidStateException
>   - Description: Popped variable is not an float: 51957
>   - In function: float &__cdecl REGoth::Scripting::DaedalusVM::popFloatReference(void)
>   - In file: G:\Source\Repos\REGoth-bs\src\scripting\daedalus\REGothDaedalusVM.cpp:533
> 
> Stack trace:
> bs::CrashHandler::reportCrash() - 0x140733526877018 File[bswin32crashhandler.cpp:473 (0)] Module[bsf.dll]
> REGoth::Scripting::DaedalusVM::popFloatReference() - 0x140696193755893 File[regothdaedalusvm.cpp:533 (117)] Module[REGothWorldViewer.exe]
> REGoth::Scripting::DaedalusVM::executeInstructionAtPC() - 0x140696193748215 File[regothdaedalusvm.cpp:262 (8)] Module[REGothWorldViewer.exe]
> REGoth::Scripting::DaedalusVM::executeUntilReturn() - 0x140696193750648 File[regothdaedalusvm.cpp:54 (0)] Module[REGothWorldViewer.exe]
> REGoth::Scripting::DaedalusVMForGameWorld::runStateLoopFunction() - 0x140696193496755 File[daedalusvmforgameworld.cpp:117 (0)] Module[REGothWorldViewer.exe]
> REGoth::AI::ScriptState::doAIState() - 0x140696193800965 File[scriptstate.cpp:295 (55)] Module[REGothWorldViewer.exe]
> bs::SceneManager::_fixedUpdate() - 0x140733528946921 File[bsscenemanager.cpp:557 (24)] Module[bsf.dll]
> bs::CoreApplication::runMainLoop() - 0x140733528853760 File[bscoreapplication.cpp:240 (33)] Module[bsf.dll]
> REGoth::REGothEngine::run() - 0x140696193334206 File[regothengine.cpp:195 (0)] Module[REGothWorldViewer.exe]
> REGoth::main() - 0x140696193333682 File[regothengine.cpp:262 (0)] Module[REGothWorldViewer.exe]
> main() - 0x140696193323677 File[main_worldviewer.cpp:93 (15)] Module[REGothWorldViewer.exe]
> __scrt_common_main_seh() - 0x140696194826372 File[exe_common.inl:288 (34)] Module[REGothWorldViewer.exe]
> 0x140734945982836 Module[KERNEL32.DLL]
> 0x140734950908529 Module[ntdll.dll]

The Host character class name is `C_NPC` and instance name is `KDW_14000_ADDON_SATURAS_ADW`.

mworchel:
Running it with G1 and `OLDMINE.ZEN` also crashes for me. Again the old error `Model Script seems to not have any meshes!` for `HUMANS`. Did you forget to commit something?

ataulien:
Can you update the bszenlib submodule?

mworchel:
All submodules are up to date. Still getting the error for `OLDMINE.ZEN`

mworchel:
Could this be related?

> G:\Source\Repos\REGoth-bs\lib\BsZenLib\lib\ZenLib\zenload\modelScriptParser.cpp(815):[enum ZenLoad::ModelScriptTextParser::Result __cdecl ZenLoad::ModelScriptTextParser::parsePfxEvent(void)] Parsed malformed PFX Start Event
> G:\Source\Repos\REGoth-bs\lib\BsZenLib\lib\ZenLib\zenload\modelScriptParser.cpp(811):[enum ZenLoad::ModelScriptTextParser::Result __cdecl ZenLoad::ModelScriptTextParser::parsePfxEvent(void)] Unknown string at the end of PFX Start Event BIP01 HEAD
> G:\Source\Repos\REGoth-bs\lib\BsZenLib\lib\ZenLib\zenload\modelScriptParser.cpp(811):[enum ZenLoad::ModelScriptTextParser::Result __cdecl ZenLoad::ModelScriptTextParser::parsePfxEvent(void)] Unknown string at the end of PFX Start Event BIP01 HEAD
> G:\Source\Repos\REGoth-bs\lib\BsZenLib\lib\ZenLib\zenload\modelScriptParser.cpp(811):[enum ZenLoad::ModelScriptTextParser::Result __cdecl ZenLoad::ModelScriptTextParser::parsePfxEvent(void)] Unknown string at the end of PFX Start Event ZS_RIGHTHAND
> G:\Source\Repos\REGoth-bs\lib\BsZenLib\lib\ZenLib\zenload\modelScriptParser.cpp(811):[enum ZenLoad::ModelScriptTextParser::Result __cdecl ZenLoad::ModelScriptTextParser::parsePfxEvent(void)] Unknown string at the end of PFX Start Event ZS_RIGHTHAND
> G:\Source\Repos\REGoth-bs\lib\BsZenLib\lib\ZenLib\zenload\modelScriptParser.cpp(811):[enum ZenLoad::ModelScriptTextParser::Result __cdecl ZenLoad::ModelScriptTextParser::parsePfxEvent(void)] Unknown string at the end of PFX Start Event ZS_RIGHTHAND
> G:\Source\Repos\REGoth-bs\lib\BsZenLib\lib\ZenLib\zenload\modelScriptParser.cpp(811):[enum ZenLoad::ModelScriptTextParser::Result __cdecl ZenLoad::ModelScriptTextParser::parsePfxEvent(void)] Unknown string at the end of PFX Start Event ZS_RIGHTHAND
> G:\Source\Repos\REGoth-bs\lib\BsZenLib\lib\ZenLib\zenload\modelScriptParser.cpp(811):[enum ZenLoad::ModelScriptTextParser::Result __cdecl ZenLoad::ModelScriptTextParser::parsePfxEvent(void)] Unknown string at the end of PFX Start Event BIP01 HEAD
> G:\Source\Repos\REGoth-bs\lib\BsZenLib\lib\ZenLib\zenload\modelScriptParser.cpp(811):[enum ZenLoad::ModelScriptTextParser::Result __cdecl ZenLoad::ModelScriptTextParser::parsePfxEvent(void)] Unknown string at the end of PFX Start Event BIP01 HEAD
> G:\Source\Repos\REGoth-bs\lib\BsZenLib\lib\ZenLib\zenload\modelScriptParser.cpp(811):[enum ZenLoad::ModelScriptTextParser::Result __cdecl ZenLoad::ModelScriptTextParser::parsePfxEvent(void)] Unknown string at the end of PFX Start Event BIP01 HEAD
> G:\Source\Repos\REGoth-bs\lib\BsZenLib\lib\ZenLib\zenload\modelScriptParser.cpp(811):[enum ZenLoad::ModelScriptTextParser::Result __cdecl ZenLoad::ModelScriptTextParser::parsePfxEvent(void)] Unknown string at the end of PFX Start Event BIP01 HEAD
> G:\Source\Repos\REGoth-bs\lib\BsZenLib\lib\ZenLib\zenload\modelScriptParser.cpp(856):[enum ZenLoad::ModelScriptTextParser::Result __cdecl ZenLoad::ModelScriptTextParser::parseSfxEvent(void)] MODELSCRIPT: Invalid eventSFX-Option: EMTPY_SLOT
> G:\Source\Repos\REGoth-bs\lib\BsZenLib\lib\ZenLib\zenload\modelScriptParser.cpp(856):[enum ZenLoad::ModelScriptTextParser::Result __cdecl ZenLoad::ModelScriptTextParser::parseSfxEvent(void)] MODELSCRIPT: Invalid eventSFX-Option: EMTPY_SLOT
> G:\Source\Repos\REGoth-bs\lib\BsZenLib\lib\ZenLib\zenload\modelScriptParser.cpp(856):[enum ZenLoad::ModelScriptTextParser::Result __cdecl ZenLoad::ModelScriptTextParser::parseSfxEvent(void)] MODELSCRIPT: Invalid eventSFX-Option: EMTPY_SLOT
> G:\Source\Repos\REGoth-bs\lib\BsZenLib\lib\ZenLib\zenload\modelScriptParser.cpp(856):[enum ZenLoad::ModelScriptTextParser::Result __cdecl ZenLoad::ModelScriptTextParser::parseSfxEvent(void)] MODELSCRIPT: Invalid eventSFX-Option: EMTPY_SLOT
> G:\Source\Repos\REGoth-bs\lib\BsZenLib\lib\ZenLib\zenload\modelScriptParser.cpp(811):[enum ZenLoad::ModelScriptTextParser::Result __cdecl ZenLoad::ModelScriptTextParser::parsePfxEvent(void)] Unknown string at the end of PFX Start Event ZS_RIGHTHAND
> G:\Source\Repos\REGoth-bs\lib\BsZenLib\lib\ZenLib\zenload\modelScriptParser.cpp(811):[enum ZenLoad::ModelScriptTextParser::Result __cdecl ZenLoad::ModelScriptTextParser::parsePfxEvent(void)] Unknown string at the end of PFX Start Event Bip01 R Hand
> G:\Source\Repos\REGoth-bs\lib\BsZenLib\lib\ZenLib\zenload\modelScriptParser.cpp(811):[enum ZenLoad::ModelScriptTextParser::Result __cdecl ZenLoad::ModelScriptTextParser::parsePfxEvent(void)] Unknown string at the end of PFX Start Event Bip01 L Hand
> G:\Source\Repos\REGoth-bs\lib\BsZenLib\lib\ZenLib\zenload\modelScriptParser.cpp(811):[enum ZenLoad::ModelScriptTextParser::Result __cdecl ZenLoad::ModelScriptTextParser::parsePfxEvent(void)] Unknown string at the end of PFX Start Event BIP01
> G:\Source\Repos\REGoth-bs\lib\BsZenLib\lib\ZenLib\zenload\modelScriptParser.cpp(811):[enum ZenLoad::ModelScriptTextParser::Result __cdecl ZenLoad::ModelScriptTextParser::parsePfxEvent(void)] Unknown string at the end of PFX Start Event zs_righthand
> G:\Source\Repos\REGoth-bs\lib\BsZenLib\lib\ZenLib\zenload\modelScriptParser.cpp(811):[enum ZenLoad::ModelScriptTextParser::Result __cdecl ZenLoad::ModelScriptTextParser::parsePfxEvent(void)] Unknown string at the end of PFX Start Event zs_righthand
> G:\Source\Repos\REGoth-bs\lib\BsZenLib\lib\ZenLib\zenload\modelScriptParser.cpp(811):[enum ZenLoad::ModelScriptTextParser::Result __cdecl ZenLoad::ModelScriptTextParser::parsePfxEvent(void)] Unknown string at the end of PFX Start Event Bip01 L Hand
> G:\Source\Repos\REGoth-bs\lib\BsZenLib\lib\ZenLib\zenload\modelScriptParser.cpp(811):[enum ZenLoad::ModelScriptTextParser::Result __cdecl ZenLoad::ModelScriptTextParser::parsePfxEvent(void)] Unknown string at the end of PFX Start Event Bip01 L Forearm
> G:\Source\Repos\REGoth-bs\lib\BsZenLib\lib\ZenLib\zenload\modelScriptParser.cpp(811):[enum ZenLoad::ModelScriptTextParser::Result __cdecl ZenLoad::ModelScriptTextParser::parsePfxEvent(void)] Unknown string at the end of PFX Start Event Bip01 L Hand
> G:\Source\Repos\REGoth-bs\lib\BsZenLib\lib\ZenLib\zenload\modelScriptParser.cpp(811):[enum ZenLoad::ModelScriptTextParser::Result __cdecl ZenLoad::ModelScriptTextParser::parsePfxEvent(void)] Unknown string at the end of PFX Start Event Bip01 L Hand
> G:\Source\Repos\REGoth-bs\lib\BsZenLib\lib\ZenLib\zenload\modelScriptParser.cpp(811):[enum ZenLoad::ModelScriptTextParser::Result __cdecl ZenLoad::ModelScriptTextParser::parsePfxEvent(void)] Unknown string at the end of PFX Start Event Bip01 R Hand
> [DEBUG] Caching Animation: HUMANS-S_WALK
> ...

ataulien:
No, that happens for me as well. I'll check whether I forgot to push when I get home.

ataulien:
> Thanks for the hint with bs::f. Say, is this supposed to successfully run all worlds with init scripts, at least without crashing? When loading ADDONWORLD.ZEN I get an exception

I can reproduce this. Let me see if I can figure it out..

src/AI/Pathfinder.cpp
```diff
@@ -0,0 +1,448 @@
...
+    bs::Vector3 Pathfinder::getTargetEntityPosition() const
+    {
+      if (!isTargetAnEntity()) return bs::Vector3(0, 0, 0);
+
+      return mActiveRoute.targetEntity->getTransform().pos();
+    }
+
+    bool Pathfinder::isTargetAnEntity() const
+    {
+      return !!mActiveRoute.targetEntity;
```
mworchel: Double negation?

src/AI/ScriptState.cpp
```diff
@@ -0,0 +1,591 @@
...
+
+    void ScriptState::requestEndActiveState()
+    {
+      // Let the state exit normally over the next cycles
+      mCurrentState.phase = AIState::Phase::End;
+    }
+
+    void ScriptState::startNativeAIState(NativeState state)
+    {
+      bs::UINT32 index = (bs::UINT32)state;
```
mworchel: May want to use C++ style cast here?

ataulien: IMHO C-Style casts are nicer for simple casts between integer types as they don't bring so much bloat with them.

If feel

```cpp
for (bs::UINT32 i = 0; i < (bs::UINT32)vector.size(); i++) { ... }
```

is more readable than

```cpp
for (bs::UINT32 i = 0; i < static_cast<UINT32>(vector.size()); i++) { ... }
```

for example. I do agree that one should use `reinterpret_cast` to do pointer casting between objects though. But it might just be that I'm not as used to C++ style casts.

src/AI/Pathfinder.cpp
ataulien: That's a way to convert something to a boolean-type. I'll change it to something more expected.

src/AI/ScriptState.cpp
mworchel: I'm fine with that. I always use C++ style casts for everything as I find them to carry more information about the intent. But in the end it usually comes down to personal taste.

src/AI/Pathfinder.cpp
mworchel: Ah I thought `targetEntity` is already boolean. Usually `return mActiveRoute.targetEntity` should work aswell. I can imagine the bsf handles overload `operator bool` similar to standard smart pointers.

ataulien: It's not boolean, but there is an operator converting it to bool. I just wanted to make that more explicit.

src/RTTI/RTTIUtil.hpp
```diff
@@ -37,11 +38,32 @@
...
+ **/
+#define REGOTH_DECLARE_RTTI_FOR_REFLECTABLE(classname)           \
+public:                                                          \
   friend class RTTI_##classname;                                 \
   static bs::RTTITypeBase* getRTTIStatic();                      \
-  decltype(classname::getRTTIStatic()) getRTTI() const override; 
+  decltype(classname::getRTTIStatic()) getRTTI() const override; \
+                                                                 \
+private:                                                         \
+  static bs::SPtr<classname> rttiCreateEmpty()                   \
```
mworchel: Better make this inline, otherwise it could (?) violate the one definition rule.

src/RTTI/RTTI_EventMessages.hpp
```diff
@@ -0,0 +1,157 @@
+#include <AI/EventMessage.hpp>
+#include <RTTI/RTTIUtil.hpp>
+
+namespace REGoth
+{
+  namespace AI
+  {
+    using UINT32 = bs::UINT32;
```
mworchel: Didn't think of this idea to circumvent `using namespace bs` :D We should definitely open a PR to the bs::f repo in the future to really fix the issue.

src/RTTI/RTTI_Pathfinder.hpp
```diff
@@ -0,0 +1,36 @@
...
+
+    public:
+      RTTI_Pathfinder()
+      {
+      }
+
+      REGOTH_IMPLEMENT_RTTI_CLASS_FOR_REFLECTABLE(Pathfinder)
+    };
+  }  // namespace AI
+  // namespace AI
```
mworchel: Extra comment

src/RTTI/RTTI_ScriptState.hpp
```diff
@@ -0,0 +1,83 @@
...
+
+    public:
+      RTTI_ScriptState()
+      {
+      }
+
+      REGOTH_IMPLEMENT_RTTI_CLASS_FOR_REFLECTABLE(ScriptState)
+    };
+  }  // namespace AI
+  // namespace AI
```
mworchel: Again extra comment

src/components/Character.cpp
```diff
@@ -1,28 +1,83 @@
...
+    : ScriptBackedBy(parent, "C_NPC", instance, gameWorld)
   {
     setName("Character");
   }
 
   void Character::onInitialized()
   {
-    ScriptBackedBy::onInitialized();
+    // Only run this when we're not getting deserialized
+    if (!hasInstantiatedScriptObject())
```
mworchel: Since I've had to deal with similar checks, I wonder if there is a canonical bsf way of checking if we are being deserialized?

src/components/Character.hpp
```diff
@@ -56,7 +86,38 @@ namespace REGoth
...
+     *         automatically start after spawning.
+     *         On Humans, this will return an empty string.
+     */
+    const bs::String& getStartAIState();
+
+    /**
+     * @return Name of the script function to call for the daily routine.
+     */
+    bs::String dailyRoutine();
+    void setDailyRoutine(const bs::String& newDailyRoutine);
```
mworchel: Could you add a small comment to this function? In particular what `newDailyRoutine` might be.

src/components/EventQueue.cpp
```diff
@@ -0,0 +1,158 @@
...
+      {
+        it = mEventQueue.erase(it);
+      }
+      else
+      {
+        it++;
+      }
+    }
+
+    if (mEventQueue.empty()) return;
```
mworchel: This return is not required as the following for loop would be a no-op anyway if the queue is empty.

src/components/Waynet.cpp
```diff
@@ -43,6 +50,274 @@ namespace REGoth
...
+      {
+        nearestDistance = thisDistance;
+        nearestIndex    = i;
+      }
+    }
+
+    return mWaypoints[nearestIndex];
+  }
+
+  HWaypoint Waynet::findSecondClosestWaypointTo(const bs::Vector3& position)
```
mworchel: Since we have to iterate over all waypoints anyway we could probably merge `findClosestWaypointTo` and `findSecondClosestWaypointTo` into one function. The simplest solution would be creating a list of waypoints sorted by distance but that is O(n) in space. A function that returns the closest and second closest waypoint is probably better. If one gets really fancy one could possibly even use templates to get something like
```cpp
template<unsigned int N>
getNClosestWaypointsTo()
```
Just some ideas.

src/components/Waynet.hpp
```diff
@@ -47,12 +50,83 @@ namespace REGoth
...
+
+    /**
+     * Finds a way between two waypoints in the given waypoint instance
+     *
+     * @return List of all waypoints that need to be visited. Will be empty if none was found.
+     */
+    bs::Vector<HWaypoint> findWay(HWaypoint from, HWaypoint to);
+
+    /**
+     * Registers the given waypoint in the waynet.
```
mworchel: Actually, it registers the given *freepoint* in the waynet.

src/scripting/daedalus/DaedalusVMForGameWorld.cpp
```diff
@@ -196,11 +440,35 @@ namespace REGoth
       mStack.pushInstance((SymbolIndex)symbolIndex);
     }
 
+    void DaedalusVMForGameWorld::external_HLP_IsValidNpc()
+    {
+      HCharacter character = popCharacterInstance();
+
+      mStack.pushInt(!!character ? 1 : 0);
```
mworchel: Sorry to pick on this again but why not just do `mStack.pushInt(character ? 1 : 0);`? `bs::GameObjectHandle` allows *direct* conversion to bool.

mworchel:
It compiles just fine and I tested it with OLDMINE.ZEN, WORLD.ZEN and NEWWORLD.ZEN. Are characters supposed to follow their daily routines yet or did you just implement the foundation? Aside from the few nitpicks the code looks fine (although I couldn't cross-check everything in absolute detail).

ataulien:
I'm not home yet so I can't fix what you found until tomorrow.

Right now characters are moving to their routines target location for me. I made most of them spawn on their routines first entry anyway, so they just stand there. I'd like to keep it like that for this PR though, as it contains most of the boiler plate code and works okay-ish.

src/components/Character.cpp
ataulien: I don't know. Haven't seen something like that yet, but I assume there must indeed be a better way.

src/RTTI/RTTIUtil.hpp
ataulien: The linker would complain if you were to define the same static function twice. Do you have an example on what could go wrong?

src/RTTI/RTTI_Pathfinder.hpp
ataulien: Thank you, clang-format! :+1:

src/components/Character.hpp
ataulien: Of course. Most of these methods are just stubs and I think many of them would be better off in the AI component. I just dumped everything gothic had in it's original engine which was a script external and started with NPC_ here.

I think it's a flaw in the original design which they had to make because they had the event queue also inside the npc class. We have a separate component for it and can clean this up a bit.

src/components/Waynet.cpp
ataulien: I'm not sold on throwing templates at that problem. But getting both nearest and second nearest in one go is a nice efficient idea!

src/scripting/daedalus/DaedalusVMForGameWorld.cpp
ataulien: I think the direct conversion to bool is fine in if statements, but it just looks weird in these cases. I'll refactor it to an explicit `if`, if you don't mind.

mworchel: The explicit `if` looks even more strange :D. Like
```cpp
if(true)
{
  return true;
}
else
{
  return false;
}
```

With stl smart pointers you often do exactly this:
```
if(pointer) ...

pointer ? ... : ...
```
It's just the short way of writing `pointer == nullptr` and I think the intent is pretty much clear.

But if you don't want to do it that way, better keep the double negation because that somewhat even looks cleaner than the explicit way.

src/RTTI/RTTIUtil.hpp
mworchel: I was just wondering if anything could go wrong by *defining* a non-inline member function in the header and the header being included in multiple cpps. But probably not, as the compiler doesn't seem to care in a minimal example.

src/components/Waynet.cpp
mworchel: It's probably an overkill as long as we don't need more than the second closest point. You could also try to merge it with the freepoint finding logic. Btw: Does it mention in the code what a freepoint is exactly (haven't looked in the wiki yet)?

src/scripting/daedalus/DaedalusVMForGameWorld.cpp
ataulien: I agree your first example looks weird. But remember that we don't check a constant `true`, which is some part of the weirdness. Also, we don't return `true` nor `false` but we call a method to push something on the stack with the intefers `1` or `0`. I assume that makes it look a lot less weird.

src/RTTI/RTTIUtil.hpp
ataulien: Ah, now I know what you mean. For classes this is indeed a bit weird, I agree.

src/components/Character.cpp
ataulien: After checking some bsf-components, it seems like they also check whether some internal state is already there before initializing...

PR #36, opened 2019-05-29
Sky color

ataulien:
Ports some of the Sky-Logic from the old REGoth repository. This PR adds a component to the `GameWorld`, which calculates the sky color depending on the daytime using the Gothic 2 sky colors.

To test this, one can modify `GameClock` from this:

```cpp
constexpr float CLOCK_SPEED_FACTOR  = SECONDS_IN_A_DAY/6000;
``` 
to

```cpp
constexpr float CLOCK_SPEED_FACTOR  = SECONDS_IN_A_DAY/60;
``` 
for 60 second days.

[WIP]: Color is not applied to any meshes now, only the clear color is modified.
Using an indoor-world like `OLDMINE.ZEN` works for testing, but you need to peek through the wall with your camera to see the effect.

Also #35 should be merged first.

### DIFF
--- a/src/components/GameClock.cpp
+++ b/src/components/GameClock.cpp
@@ -1,24 +1,27 @@
 #include "GameClock.hpp"
+#include <Math/BsMath.h>
 #include <RTTI/RTTI_GameClock.hpp>
 #include <Scene/BsSceneObject.h>
-#include <Math/BsMath.h>
 
 namespace REGoth
 {
   constexpr float SECONDS_IN_A_MINUTE = 60;
   constexpr float SECONDS_IN_AN_HOUR  = 60 * SECONDS_IN_A_MINUTE;
   constexpr float SECONDS_IN_A_DAY    = 24 * SECONDS_IN_AN_HOUR;
-  constexpr float CLOCK_SPEED_FACTOR  = SECONDS_IN_A_DAY/6000; // see https://forum.worldofplayers.de/forum/threads/396326-Tipp-Tageweise-springen-%28auch-zur%C3%BCck%29?p=6231841&viewfull=1#post6231841
+  // see
+  // https://forum.worldofplayers.de/forum/threads/396326?p=6231841&viewfull=1#post6231841
+  constexpr float CLOCK_SPEED_FACTOR = SECONDS_IN_A_DAY / 6000;
 
   GameClock::GameClock(const bs::HSceneObject& parent)
-    : bs::Component(parent)
-  { }
+      : bs::Component(parent)
+  {
+  }
 
   void GameClock::fixedUpdate()
   {
     float delta = bs::gTime().getFixedFrameDelta();
 
-    mElapsedSeconds       += delta;
+    mElapsedSeconds += delta;
     mElapsedIngameSeconds += delta * CLOCK_SPEED_FACTOR;
   }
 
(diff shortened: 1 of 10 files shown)


src/components/GameClock.cpp
```diff
@@ -1,24 +1,28 @@
...
 namespace REGoth
 {
   constexpr float SECONDS_IN_A_MINUTE = 60;
   constexpr float SECONDS_IN_AN_HOUR  = 60 * SECONDS_IN_A_MINUTE;
   constexpr float SECONDS_IN_A_DAY    = 24 * SECONDS_IN_AN_HOUR;
-  constexpr float CLOCK_SPEED_FACTOR  = SECONDS_IN_A_DAY/6000; // see https://forum.worldofplayers.de/forum/threads/396326-Tipp-Tageweise-springen-%28auch-zur%C3%BCck%29?p=6231841&viewfull=1#post6231841
+  // see
+  // https://forum.worldofplayers.de/forum/threads/396326-Tipp-Tageweise-springen-%28auch-zur%C3%BCck%29?p=6231841&viewfull=1#post6231841
+  // constexpr float CLOCK_SPEED_FACTOR = SECONDS_IN_A_DAY / 6000;
+  constexpr float CLOCK_SPEED_FACTOR = SECONDS_IN_A_DAY / 60;
```
mworchel: I think you didn't intend to commit this change?

src/components/GameClock.cpp
```diff
@@ -30,80 +34,95 @@ namespace REGoth
...
-    float elapsedSecondsOfHour = mElapsedIngameSeconds - (getDay() * SECONDS_IN_A_DAY) - (getHour() * SECONDS_IN_AN_HOUR);
+    float elapsedSecondsOfHour =
+        mElapsedIngameSeconds - (getDay() * SECONDS_IN_A_DAY) - (getHour() * SECONDS_IN_AN_HOUR);
 
     return bs::Math::floorToPosInt(elapsedSecondsOfHour / SECONDS_IN_A_MINUTE);
   }
 
+  float GameClock::getDayRatio() const
+  {
+    float elapsedSecondsOfDay = mElapsedIngameSeconds - (getDay() * SECONDS_IN_A_DAY);
```
mworchel: Maybe create a function `getCurrentTimeOfDayInSeconds` or so as the exact same snippet is used in `GameClock::isTime`.

src/components/GameClock.hpp
```diff
@@ -74,13 +82,13 @@ namespace REGoth
...
-  /* RTTI                                                                 */
-  /************************************************************************/
+    /************************************************************************/
+    /* RTTI                                                                 */
+    /************************************************************************/
   public:
     REGOTH_DECLARE_RTTI(GameClock)
 
-  // protected:
+    // protected:
```
mworchel: What's up with the weird formatting (comments getting indented)? Or is it just following the rules of the format file?

ataulien: I assume Clang-format thinks the comment is documenting the next class member and thus it should be indented. I actually didn't notice that this class still uses the old workaround. Will fix it while I'm at it.

src/components/GameClock.cpp
ataulien: Ah, darn it, I didn't.

src/sky/SkyColoring.cpp
```diff
@@ -0,0 +1,369 @@
...
+    bs::UINT32 si0 = (bs::INT32)mSkyStates.size() - 1;
+    bs::UINT32 si1 = 0;
+
+    // Find the two states we're interpolating
+    for (bs::UINT32 i = 0; i < (bs::UINT32)mSkyStates.size(); i++)
+    {
+      // Interpolate between the current and the next
+      if (mMasterState.time < mSkyStates[i].time)
+      {
+        si0 = (i - 1) % mSkyStates.size();
```
mworchel: Are you sure, doing `i - 1` if `i = 0` is a good idea? Also is the intent to interpolate between the previous and the next state (that would be strange) or the current and the next state as stated in the comment?

src/sky/SkyColoring.cpp
```diff
@@ -0,0 +1,369 @@
...
+    // Note: These are some magic values to match what gothic does
+    float fogMidrange = farPlane * 0.4f;
+    float fogMidDelta = farPlane - fogMidrange;
+
+    // Do heightfog-approximation
+    float ytotal  = worldBBox.getMax().y - worldBBox.getMin().y;
+    float fogMinY = worldBBox.getMin().y + 0.5f * ytotal;
+    float fogMaxY = worldBBox.getMin().y + 0.7f * ytotal;
+
+    // Skale fog back depending on how high the camera is
```
mworchel: `Skale` ->`Scale`

src/sky/SkyColoring.cpp
```diff
@@ -0,0 +1,369 @@
...
+  bs::Color SkyColoring::getPolyCloudsLayerColor()
+  {
+    bs::Color color;
+
+    if (isNightTime())
+      color = bs::Color::White;
+    else
+      color = mMasterState.domeColorUpper;
+
+    if (mMasterState.time >= TIME_KEY_3 && mMasterState.time <= TIME_KEY_5)
```
mworchel: That this handles the late night and midnight states (without evening and dawn)

src/sky/SkyColoring.cpp
```diff
@@ -0,0 +1,369 @@
...
+    {
+      color = 0.5f * (color + bs::Color::White);
+    }
+
+    return color;
+  }
+
+}  // namespace REGoth
+// namespace REGoth
```
mworchel: Thanks again clang-format :D

mworchel:
Doesn't work for me. The sky stays blue in G1 and G2, no matter the world. It seems you forgot to commit the part where the `Sky` component is added to a scene object.

ataulien:
Hmm, I remember amending that to the commit. I can't check what went wrong currently though. If you want, you can simply add the Sky component to the game world. Nothing else is required.

mworchel:
When manually adding the component to the game world, the sky color does seem to change according to the time. Did you figure out what went wrong?

src/sky/SkyColoring.cpp
ataulien: That looks weird indeed, but it is correct.

For example: Let's say the master states time is `0.30` and we got state A at `0.00` and state B at `0.25` and state C at `0.50`. Since they are ordered, we would need to find the first state with a time larger than the current time. In this case, that would be state C. Of course we overshot by one, since we're interested in state B, that's where the subtraction comes from.
As long as they're set up correctly, the case that `i` will be 0 will not happen.

I agree that all this should be made more clear. Finding the first state with a time *less* than the current time would also work and be much more readable.

src/sky/SkyColoring.cpp
ataulien: I refactored the code. But I don't see a point into making a different component here.

src/components/GameClock.cpp
ataulien: Done!

src/sky/SkyColoring.cpp
mworchel: Thanks for the additional clarification!

Especially I didn't think of this when making my comment: 

> Subtracting 1 will not cause a negative numbers here **since the first state has a time of 0.0**

mworchel:
Works.

Issue #28, opened 2019-05-14
Automatically find Startpoint


ataulien:
Right now, the Startpoint to insert the Hero at is hardcoded inside the sample applications, like here:

https://github.com/REGoth-project/REGoth-bs/blob/0bbc257854458c77920e2dc979ded1c87bfab287/src/main_WorldViewer.cpp#L31-L39

Unfortunately, the name of the startpoint differs between ZENs, but there is only supposed to be a single one of them in each ZEN I think. As far as I know, the original engine simply searches for an object of class `zCVobStartpoint`.

The code which first touches the startpoint on import is here:
https://github.com/REGoth-project/REGoth-bs/blob/0bbc257854458c77920e2dc979ded1c87bfab287/src/main_WorldViewer.cpp#L31-L39

A simple solution would be to rename the scene object just `STARTPOINT` and then search for it by name. I'm not sure whether the actual name of the startpoint is important as there should be only one anyways. We could also pass it to the `GameWorld` somehow. Although I don' like making a setter for that.

ataulien:
Renaming the startpoint to some known name has been done in #39. However, the importer could still handle the startpoint better, e.g. by storing the handle to it and providing it via a getter.

If someone wants to take a shot at improving the importer, there are some hints in #39 of what could be done.

Issue #14, opened 2019-04-25
World-Mesh is half-imported even though it is cached


ataulien:
When importing a ZEN-file, the `packMesh` method should only be called if the world-mesh has not already been cached:

https://github.com/REGoth-project/REGoth-bs/blob/8f785ce6ffcd969bee93685236e7cac0e23fafcb/src/world/internals/ConstructFromZEN.cpp#L91-L107

To fix this, the module has to be re-structured a bit so that the packing happens after checking for the cache.

PR #39, opened 2019-06-09
Give zCVobStartpoint's scene object a well defined name

ghtyrant:
This implements a simple solution to #28.
I'm not sure if there's a better place for the WORLD_STARTPOINT constant, just let me know.

### DIFF
--- a/src/components/GameWorld.cpp
+++ b/src/components/GameWorld.cpp
@@ -16,6 +16,8 @@
 
 namespace REGoth
 {
+  const char* const WORLD_STARTPOINT = "STARTPOINT";
+
   GameWorld::GameWorld(const bs::HSceneObject& parent, const bs::String& zenFile)
       : bs::Component(parent)
       , mZenFile(zenFile)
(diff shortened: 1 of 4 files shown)


ataulien:
Thanks for your contribution!

I do have a few suggestions though:

 1. Right now, code external to the world needs to know that we fix up the name of the startpoint and which name we have given it. This is an internal detail of the import process and I'd like to keep such information away from the game code. In my opinion it would be cleaner to have the `GameWorld`-Component search the startpoint and store a reference to it. This could then be made public over a getter. 
 2. Can we get rid of the `external const char*`? I think that would be the first of those in our codebase and it seems to be a more C-ish pattern.

Most of these problems (and also #14) could be neatly solved if the world importer was a class, rather than a set of functions. (Sometimes my inner C programmer shines through :sweat_smile:) That class could simply store the startpoint and make it available to the game world. 

The waynet-object has a similar issue. It gets assigned the name `waynet` in the importer and the `GameWorld` just looks for that. I dislike how that string is not shared between the importer and the `GameWorld`-class. Maybe a header with some string-constants for objects which have special names would also do the job... I'm just rambling at this point, what do you think?

ghtyrant:
Thanks for the detailed response. I see your concerns.

I will try to wrap the importer in a class, that's the cleanest way to solve this, IMHO. This will also get rid of the need for global constants, since we can access the objects using a getter.

ataulien:
I'll merge this as it is working and is a nice thing to have. Still, I will update the corresponding issue.

PR #46, opened 2019-07-01
Dialogue ui

ataulien:
This PR implements a very simply but working dialogue UI. The interesting parts are "under the hood" however: Most underlying systems for conversations with NPCs have been implemented, as well as some externals to make a few dialogue actions work.

For example: You can talk with Diego when starting G1. After chosing END, he will switch routines and walk to the old camp.

Note that NPCs won't actually talk since `AI_Output` has not been implemented yet. That doesn't matter however, since the scripts are executed anyways.


src/components/CharacterKeyboardInput.cpp
```diff
@@ -35,6 +28,65 @@ namespace REGoth
...
+      REGOTH_THROW(
+          InvalidStateException,
+          bs::StringUtil::format("Scene Object {0} does not have a CharacterEventQueue component!",
+                                 SO()->getName()));
+    }
+  }
+
+  void CharacterKeyboardInput::update()
+  {
+    if (bs::gVirtualInput().isButtonDown(mAction))
```
KirmesBude: I think we need to think about how to handle input in non-continous cases.
This would trigger multiple times for one buttonpress (since it will be held down for more than one update cycle), no?

src/gui/skin_gothic.cpp
```diff
@@ -36,42 +30,70 @@ namespace REGoth
...
+
+    skin->setStyle("GothicSubtitleBoxCharacterName", labelDefaultHighlighted);
+    skin->setStyle("GothicSubtitleBoxText", labelDefault);
+
+    // skin->setStyle("Button", const GUIElementStyle &style)
+
+    // Cache the skin for later use
+    s_SkinGothic = skin;
+
+    return bs::BuiltinResources::instance().getGUISkin();
```
KirmesBude: Does this not get rid of all the ElementStyle definitions we do here and just returns the skin provided by bsf?

src/gui/skin_gothic.cpp
```diff
@@ -36,42 +30,70 @@ namespace REGoth
...
+    bs::GUIElementStyle labelObjectFocus = baseStyle;
+    labelObjectFocus.font                = fontGothicOld10WhiteHi;
+    labelObjectFocus.fontSize            = 17;
+    labelObjectFocus.wordWrap            = false;
+    // labelObjectFocus.textHorzAlign = bs::TextHorzAlign::THA_Center;
+    // labelObjectFocus.textVertAlign = bs::TextVertAlign::TVA_Center;
+    skin->setStyle("GothicLabelObjectFocus", labelObjectFocus);
+
+    bs::GUIElementStyle subtitleBox = baseStyle;
+    subtitleBox.normal.texture      = gOriginalGameResources().sprite("DLG_CHOICE.TGA");
```
KirmesBude: You also load it in the constructor of UISubtitleBox. 
Is it not possible to use the style there instead of explicitly loading it?

src/main_WorldViewer.cpp
```diff
@@ -79,7 +80,12 @@ class REGothWorldViewer : public REGoth::REGothEngine
 
     HCharacter hero = heroSO->getComponent<Character>();
 
+    mMainCamera->SO()->setActive(false);
+    mMainCamera->SO()->setActive(true);
```
KirmesBude: Why is this necessary?

src/original-content/OriginalGameResources.hpp
```diff
@@ -0,0 +1,110 @@
+#include <BsPrerequisites.h>
```
KirmesBude: missing #pragma once?

src/components/CharacterKeyboardInput.cpp
ataulien: No, this method is specifically for this use case. It will return true only for one single update. The `isButtonHeld` will do what you have in mind.

src/gui/skin_gothic.cpp
ataulien: Yes, my custom skin doesn't really work yet :smile:

src/main_WorldViewer.cpp
ataulien: I thought I added a comment here. There is an issue about the camera not moving and this is the workaround.

src/components/CharacterKeyboardInput.cpp
KirmesBude: True! 
I played around with this a little bit and I must have used fixedUpdate() or something, eventually moving to events eventhough just using update() would have fixed my problems :/.

src/gui/skin_gothic.cpp
KirmesBude: I guess this is obsolete, since we discard the style.

src/gui/skin_gothic.cpp
KirmesBude: I would suggest adding our styles on to the built-in skins and returning that. 
Best of both worlds :P

src/gui/skin_gothic.cpp
ataulien: I originally wanted to set the background image via the style but didn't get it to work. Do you know how that could work?

src/gui/skin_gothic.cpp
ataulien: Done that.

src/gui/skin_gothic.cpp
ataulien: By the way, the style will only be loaded once, since it's cached inside the load-method. I could rename the function to make this more obvious.

KirmesBude: Do you mean the texure is cached. Yes I did not think about that so it does not really matter. 
As for setting the background image via the style: 
I remember this(https://github.com/KirmesBude/REGoth-bs/blob/feature/console/src/components/Console/Console.cpp#L35 ) was working
Style [here](https://github.com/KirmesBude/REGoth-bs/blob/feature/console/src/gui/skin_gothic.cpp#L94)

ataulien: Ah, I thought you were talking about the multiple calls to `createSkin_Gothic`. But loading the texture is also cached.

Regarding the styles: I didn't know the `GUITexture` would use the texture given in the skin. Thanks for pointing it out!

src/components/UIFocusText.cpp
```diff
@@ -0,0 +1,110 @@
...
+
+    mLabelAboveObject = layout().addNewElement<bs::GUILabel>(bs::HString(""));
+    // mLabelAboveObject->setStyle("GothicLabelObjectFocus");
+  }
+
+  UIFocusText::~UIFocusText()
+  {
+  }
+
+  void UIFocusText::putTextAbove(HFocusable focusable, bs::String text)
```
KirmesBude: Could you go into more details on how this will work? The way I see it right now there is only one UIFocusText component that handels all the Focus draws (which makes sense since there can only be one thing in front of you to focus). 
But when would this called and why would we provide a text argument instead of using getText on the Focusable?

ataulien: Good catch. The `UIFocusText` actually existed before the `Focusable`-Component. "Back then" the method wanted a `bs::HSceneObject` and the text to display.

I later figured out I should add a component to the focusable stuff.

ataulien:
> So I compiled and tested it and it seems to work fine.
> The only problem I ran into is that when I trigger the EXIT_LATER option with diego when he(or the player? or both) are behind the first turn, the game crashes with

Ah, this issue. For some reason I forgot! :sweat_smile: This will happen on other NPCs as well, usually when they are supposed to enter some state like `ZS_SLEEP`. I'll take a look at it.

> If you dont mind, could you extend the documentation with the GUI related stuff. Most importantly what would GameUI handle? Only "overlay/realtime" GUI or also the "Menus"?

Sure. It might be even better to get rid of the global and add a session-object, which does not need to be global. Also some kind of message-system would be nice so the gameplay related components are decoupled from the UI code if they want to do something. I feel like this is out of scope for this PR and I'd like to tackle it in another one.

ataulien:
Merging, as no more problem have been found.

PR #47, opened 2019-07-06
[WIP] Movement

ataulien:
Based on #46. WIP for as long as that one is not merged, but the features implemented here should be ready for testing.

Implements some new movement features:

 - Strafe using Q/E
 - Jump using Left ALT
 - Switch to Walk-Mode using Right Shift
 - Switch to Sneak-Mode using X

Some animation features like blending between two animations are still missing which will make some animations look a bit weird.


KirmesBude:
I tested this shortly. Compiles fine and works well. Jumping does not give you any height, but I think this is intended/not planned right now. 
I am unsure about the behaviour of the original game, so here goes nothing: 

- Is backing up in sneak mode intended? 
- You can only sneak-strafe to the right for some reason. [You seem to be using the right animations though, so I dont know why this happens]
- Go into walk mode, jump and then press forward mid jump and you stuck in the jump animation until you let go of forward - I feel like I remember some variation of this bug in the original game, but I am not sure if this is circumstances it happened in the original :D. [I guess some problems with transitioning between jump and walk_forward ?] 
- This also works for walk/sneak when holding forward in the middle of the backup move
- How should changing walking/running behave out of sneaking or going from walking to sneaking and the canceling sneaking (x). Currently it always goes back to running. Should it restore the last activate walkmode? 

Can you or someone else confirm these problems?

ataulien:
I can confirm all of these issues. Most of them are related to unimplemented features from the model scripts such as playing an animation in reverse. There is also an issue with Gothic 2s strafing animations. I'll merge the PR and open Issues for the things you found.

PR #51, opened 2019-07-12
Chore: Add Debian package information

charlydelta:
I added some information which I had to find out when I compiled REGoth. It should be moved to the sphinx docs eventually, but until then I just wanted these information to be conserved.

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,7 +58,7 @@ On the first run, the *Cache* will be created in the `cache/` folder. Some sampl
 ## Get the Code
 
 ```sh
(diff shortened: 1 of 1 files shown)


ataulien:
There is a shorter form for the git command: `git clone --recursive ...`, which does the same and is a bit less verbose. Mind to change it to that?

charlydelta:
Sure, I just used it because according to [this SO question][https://stackoverflow.com/questions/3796927/how-to-git-clone-including-submodules) it seemed to be preferred over `--recursive`

ataulien:
Thank you!

PR #52, opened 2019-07-15
Incorporate cxxopts for command line argument parsing

charlydelta:
This PR incorporates the cxxopts command line parsing library and offers APIs to use a default engine configuration or specify a custom configuration.

- `REGoth::REGothEngine` does not have a default constructor anymore, needs a configuration object of type `REGoth::EngineConfig` as parameter, and is abstract now.
- `REGoth::REGothEngineDefaultConfig` is an implementation which can be used in place of the old `REGoth::REGothEngine` in cases where no custom configuration is needed.
- `REGoth::REGothEngine` should be derived and the configuration class `REGoth::EngineConfig`.
- Docs and examples updated accordingly.

~~This branch depends on `cxxopts` with an applied [patch to disable exceptions](https://github.com/jarro2783/cxxopts/pull/190) (**don't merge before upstream merged this or implemented something comparable**)~~

Edit: cxxopts did not yet accept the patch, but bsf now allows to disable exceptions


ataulien:
How are your thoughts on getting a commandline flag somewhere deep into the engine? Lets say I wanted to enable script disassembly output via the commandline or so. 

I've already said that using a global object would make sense, but on a second thought, this might not be the optimal solution after all: Suddenly there would be dependencies to that commandline module all over the place!

A cleaner way IMHO would be to give every class that would need to be configured in some way a static structure containing whatever configuration it needs. Those could then be set using the commandline flags, a config file, whatever. What do you think about that?

charlydelta:
On a related note: Initially I didn't like how it is at the moment. I didn't want to make an API passing the cxxopts reference in the first place, because this way the dependency leaks into the project and I wanted to avoid that.

My first instinct was to write an abstraction API, and with your feedback I think this is actually the way to go. I decided against implementing it, because it is quite complex and I thought I would be over-engineering. So my plan now, after your feedback, would be the following:

- Create a cxxopts-agnostic abstraction layer which offers an API to register CLI arguments of arbitrary type (given that the needed templates are provided) and query their state 
- The location is in `REGothEngine` (the CMake target, not the C++ class), and only there it is (privately) linked against
- Make an global object of that abstraction layer, which can arbitrarily be called basically everywhere
- Each `main()` would basically just use that global object to feed `argc` and `argv` in, that's just about it

Also, would be super nice to have:

- Do not only register a CLI argument, but also allow passing a lambda which functions as validator. After parsing `argc` and `argv`, the abstraction layer would call the corresponding lambda with the actual argument, and throw if the lambda returns `false`

Advantages:

- We can replace cxxopts easily, if need be
- The dependency doesn't leak into the rest of the project, or other executables 
- Basically your first suggestion with a global object, however a more sophisticated one and not only a data structure

Disadvantages:

- Over-engineering or over-thinking the problem?
- Config files not considered atm, but the abstraction could also account for that. In the end, it doesn't really matter, where the configs come from. Could also be a two-layered approach eventually, where the basics are initialised from a config file, and then overwritten by CLI arguments.

> A cleaner way IMHO would be to give every class that would need to be configured in some way a static structure containing whatever configuration it needs. Those could then be set using the commandline flags, a config file, whatever. What do you think about that?

That's also possible, but this sounds more of a *per use-case*, specifically tailored kind of solution instead of a general one (or maybe I didn't understand right?). With the approach I suggested it would be unnecessary. At least if it works as I imagine, because then you'd just feed the *references* of variables to the global object. If it passes, everything is fine (and those variables are correctly initialised by then), if not (if a lambda returned false) the executable has already thrown anyway.

frabert:
I'd like to propose a different approach: we could make so that all data parsed from the CLI is put into a `CliConfig` object or something, and the config file data is put into a `FileConfig` object. A third object, let's say `ConfigStore` is used to put the two sources together, maybe to establish some sort of precedence. E.g., even if the config files say that the game should run fullscreen, I told the CLI to run windowed and the game should follow that.

I then have to ideas: the first one is the "naive" one, each game engine component gets a reference to the `ConfigStore` and gets its data from there.
The second one, is to create an ad-hoc interface for each component, so that it may refine the config data a bit more from the store before giving it to the component.

It may be complicated, but I think it may give greater flexibility if we ever want to do automated testing or stuff, since every game component could be given a "fake" interface to edit its config without messing with files

charlydelta:
Ah, basically dependency injection, good point. But I think that what you describe is not much of a "different approach", but rather a refinement of what I explained. The `ConfigStore` is basically the abstraction layer I described.

We could just make an interface, like `ConfigStoreInterface`, which is the type of the global object. In a test scenario, we could replace this with a dummy to write whatever data we want. I just don't know, how well bsf plays with his. I think that's something, you guys can answer?

The main problem we face here is that we have a multitude of types. We can't just have a map of values, because they're all of a different type. That's why I wanted to employ an approach where references of the callees are passed to the global object, and the global object just initialises them using `cxxopts` in a real-world scenario (i.e. no test scenario). Because otherwise, we'd have to implement Variants, and bsf doesn't seem to support that. Also, with templates, we can then generally use it to initialise whatever type we want. This is already supported by `cxxopts`, see for example [this](https://github.com/REGoth-project/REGoth-bs/pull/52/files#diff-977b2f32a90218a169ba6d26a39b0041R25) is the template to initialise a C++ variable of type `bs::Path` from the command line like this: 

```cpp
options.add_options()
    ("a,game-assets", "Path to a Gothic or Gothic 2 installation", cxxopts::value<bs::Path>(gameDirectory), "[PATH]")
    ;
```

frabert:
I'm sorry, I'm afraid I don't understand what you mean with "we'd need variants"...

Here's how I envision it:

````c++
struct CliConfig {
  std::string gameDirectory;
  int xRes, yRes;
  bool windowed;
  // ...
};
````
During game startup, we use `cxxopt` to populate that struct, then we use a similar method to populate the equivalent struct from the config files, and then both get fed to the `ConfigStore` as explained above.

Maybe I'm just missing the big picture here, but I don't think we need variants, because once you have parsed something, it either _is_ the correct type, or it is not. In which case, you just display an error or something

**Comment by charlydelta** [comment]
This is what I see as a big issue. Correct me if I am wrong, but we do not have one big, monolithic, static data structure which is valid for all executables and all command line flags they may have. For example, let's say we have an executable to test/debug one single feature X, which needs X as CLI flag. If we had this one data structure, we'd need to add the flag for X for *all* executables, even though only one minor executable actually needs it. That's why I wanted to make a dynamic API, where the engine (the class inheriting from `REGothEngine`) actually calls a global object and actively registers, what options it wants. 

Is this assumption wrong? Because this is how I understood the problem. A practical example would be the `--world` flag, which is actually only needed in `REGothWorldViewer`. This would just bloat the data structure for all others. I mean, it would work, sure, but I don't think this is an elegant solution.

**Comment by frabert** [comment]
Ohhh I see what you mean now! On the other hand, though, command line arguments are inherently application-specific, so maybe it doesn't make sense to have a shared interface for every program? It may make sense to make an object for the config options that are common to all programs because they are used by the engine, but every additional app-specific option should be placed in another object

**Comment by charlydelta** [comment]
I see your point. I will think about it. Thank you very much for your feedback!

File: src/REGothEngine.hpp [hunk 1]
```diff
@@ -3,14 +3,69 @@
...
+    }
+
+    virtual void registerCLIOptions(cxxopts::Options& /* options */)
+    {
+      // pass
+    }
+
+    unsigned int verbosity = 0;
+    bs::Path gameExecutable;
+    bs::Path assetsPath;
```
ataulien: Might as well call this `originalAssetsPath` or `gothicGameRoot` to not collide with REGoths own `content`-directory.

src/REGothEngine.hpp
```diff
@@ -3,14 +3,69 @@
...
+      options.parse_positional({"game-assets"});
+    }
+
+    virtual void registerCLIOptions(cxxopts::Options& /* options */)
+    {
+      // pass
+    }
+
+    unsigned int verbosity = 0;
+    bs::Path gameExecutable;
```
ataulien: Maybe call this so it's obvious that the REGoth executable is meant here? Something like `runningExecutablePath` or similar?

src/REGothEngine.hpp
```diff
@@ -3,14 +3,69 @@
...
+    bs::Path assetsPath;
+    unsigned int resolutionX = 1280;
+    unsigned int resolutionY = 768;
+    bool fullscreen = false;
```
ataulien: Boolean values should be prefixed with something that makes it obvious it's a boolean, such as `isFullscreen` or `shouldGoIntoFullscreen`.

src/main_WorldViewer.cpp
```diff
@@ -14,9 +14,27 @@
...
 
+class Config : public REGoth::EngineConfig
+{
+
+public:
+
+  virtual void registerCLIOptions(cxxopts::Options& opts) override
+  {
+    opts.add_options()
+      ("w,world", "World name to load", cxxopts::value<bs::String>(world), "[NAME]")
```
ataulien: We're not loading a "World name" here, are we? :wink: 

How about `Name of the world to load`?

src/main_WorldViewer.cpp
```diff
@@ -44,25 +62,26 @@ class REGothWorldViewer : public REGoth::REGothEngine
...
+    if (mConfig.world.empty())
     {
       REGOTH_THROW(InvalidStateException, "World cannot be empty.");
     }
 
-    bs::StringUtil::toUpperCase(mWorld);
-    if (not bs::StringUtil::endsWith(mWorld, ".ZEN"))
+    bs::String worldName = mConfig.world;
+    bs::StringUtil::toUpperCase(worldName);
+    if (not bs::StringUtil::endsWith(worldName, ".ZEN"))
```
ataulien: Could we somehow put transformations like this into the config class so the game code doesn't have to deal with it?

charlydelta:
> I like the new system. The reviews I left are only some nitpicks wink

Thank you very much for the quick review, I really appreciate that. I agree in all points and will update my changes accordingly.

> By the way, can you update the examples in the documentation with the new templates? You don't have to compile the docs if it's too much of a hassle to install everything. I can do that once the PR is merged.

No it's fine, I already worked with sphinx etc. I will do that, but only after you agree with the code. I think that there still might be changes incoming.

By the way, regarding your earlier comment:

>How are your thoughts on getting a commandline flag somewhere deep into the engine? Lets say I wanted to enable script disassembly output via the commandline or so.

I hope I understood you right and this requirement is satisfied now, isn't it? The Engine should know, what needs to be configured.

It is quite a drastic change to require this config data struct as mandatory constructor parameter, but I think that it is a reasonable amount of overhead for the gain. I.e. testability by feeding dummy configs, and the ability to later also add config file support, as pointed out by @frabert, and of course the CLI support, which was the original scope of this PR.

src/REGothEngine.hpp
charlydelta: Renamed to `originalAssetsPath`

src/REGothEngine.hpp
charlydelta: Renamed to `engineExecutablePath` (it was called like this in the original code, and I think this is the best fitting. Let me know if you disagree or think `running...` is better :)

src/REGothEngine.hpp
charlydelta: Right! `isFullscreen` it is

src/main_WorldViewer.cpp
charlydelta: Changed to `Name of the world to load`. Brain fart

src/main_WorldViewer.cpp
charlydelta: I've added a neat `verifyCLIOptions` method which will be called as last step after the parameter parsing, where the config object ca be checked for integrity (and throw if not) and inputs can be sanitised. I use this to make the engine path and the original assets path absolute now as well.

charlydelta:
This is still WIP, as documentation is still missing and I was not yet able to properly test this.

src/main.cpp
```diff
@@ -2,209 +2,13 @@
...
-    gDebug().logDebug("Set CullDistance to: " + toString(value));
-
-    sceneCamera->setRenderSettings(rs);
-  });
-
-  regoth.run();
-
-  return 0;
+  return ::REGoth::runEngine(engine);
 }
```
charlydelta: @ataulien I also removed the legacy code right here. I think it would be the best, if we make `REGoth` the main executable to actually just play. This means that this main.cpp is very small, as it just has to instantiate the plain defaults to get started.

All other executables are pretty much development, single purpose execs which modify the default engine to make a certain trait better testable/debuggable. Please let me know if this is okay or if I misunderstood anything.

ataulien: I have some code laying around defining a `REGothGame` class inherited from `REGothEngine` which sets up normal gameplay. That way we can have a simple `main.cpp` and also define test executables extending the game class if we want to.

src/REGothEngine.hpp
```diff
@@ -42,27 +44,42 @@ namespace REGoth
 
       // Define engine options
       options.add_options()
-        ("a,game-assets", "Path to a Gothic or Gothic 2 installation", cxxopts::value<bs::Path>(assetsPath), "[PATH]")
+        ("a,game-assets", "Path to a Gothic or Gothic 2 installation", cxxopts::value<bs::Path>(originalAssetsPath), "[PATH]")
```
ataulien: Historically, the flag to specify the game root has always been `-g`. I think we should keep that!

src/REGothEngine.hpp
```diff
@@ -42,27 +44,42 @@ namespace REGoth
...
       options.parse_positional({"game-assets"});
     }
 
+    void verifyCLIEngineOptions()
+    {
+      // Resolve paths.
+      engineExecutablePath.makeAbsolute(bs::FileSystem::getWorkingDirectoryPath());
+      originalAssetsPath.makeAbsolute(bs::FileSystem::getWorkingDirectoryPath());
+
+      // TODO: Verify that assetsPath is a valid Gothic or Gothic 2 installation.
```
ataulien: There are methods doing exactly those two checks:

https://github.com/REGoth-project/REGoth-bs/blob/569d3fb1b5e4d842720da22ecf9bdeed1081840b/src/original-content/OriginalGameFiles.hpp#L75-L89

ataulien: Just below that function, there is a `isGameRoot` function which is protected but could be made public.

src/REGothEngine.hpp
```diff
@@ -177,6 +194,13 @@ namespace REGoth
      */
     void shutdown();
 
+    /**
+     * Gets the engine's configuration.
+     *
+     * @return Engine configuration data structure.
+     */
+    virtual const EngineConfig& config();
```
ataulien: Function itself should also be const.

src/main_WorldViewer.cpp
```diff
@@ -22,18 +22,39 @@ class Config : public REGoth::EngineConfig
...
+  {
+    // Make sure that world is set.
+    if (world.empty())
+    {
+      REGOTH_THROW(InvalidStateException, "World cannot be empty.");
+    }
+
+    // Unify possible inputs.
+    bs::StringUtil::toUpperCase(world);
+    if (not bs::StringUtil::endsWith(world, ".ZEN"))
```
ataulien: Since when does `not` work? :open_mouth:  Is the standard c++ and good practice? Reads nicely, but it does raise eyebrows...

frabert: @ataulien https://en.wikipedia.org/wiki/C_alternative_tokens :)

mworchel: @frabert

Standard compliant: Yes. But not supported by MSVC the last time I checked.

frabert: @mworchel I just checked with VS2017 and I didn't need to include any header, FWIW

I will abstain from giving a judgment on the style, as I am quite uncertain about it, but it is at least working.

mworchel: Hm I think the last time I encoutered it was with vs2015 in some uni assignment template and I remember I had to replace all occurences of 'and' and 'or' for it to work. I would probably use '!' just to be consistent with the style of this project and many others but it's not my decision :D

charlydelta: I've always done this because I think it's way more readable, and I like writing *human readable* code. I've never touched a Microsoft compiler though, I didn't know it doesn't work there, but I do know that it is standard conform.

Doesn't matter in the end, I will of course change it to fit the general code style. I just wrote it out of habit (I also do this with `and`, `xor`, `or`, ... instead of `&&`, `^`, `||`, ... :D )

src/REGothEngine.hpp
charlydelta: Oh good point, thank you

src/main.cpp
charlydelta: Is it not feasible to try to make `REGothEngine` the default near-production implementation and override certain aspects of the engine if-need-be for tests/debugs? 

But you're saying that this change is legitimate, right? I mean, the legacy code is in Git anyway, so why keep it in the code base, no?

src/REGothEngine.hpp
charlydelta: Great, didn't know that. Will do

src/REGothEngine.hpp
charlydelta: Oh I don't have these changes yet, thanks for the heads up :). I will now commit the other fixes and do a quick rebase.

src/REGothEngine.cpp
```diff
@@ -214,60 +226,101 @@ void REGothEngine::shutdown()
...
+  bool help;
+  bool version;
+
+  cxxopts::Options options{argv[0], "REGoth - zEngine Reimplementation."};
+
+  // Add general options.
+  options.add_options()
+    ("h,help", "Print this help message", cxxopts::value<bool>(help))
+    ("version", "Print the REGoth version", cxxopts::value<bool>(version))
+    ("v,verbosity", "Verbosity level", cxxopts::value<bool>())
```
charlydelta: I'm afraid this doesn't make sense. I thought that it would be cool if we could increase the verbosity level like in some UNIX programs, e.g. `-vv` is less verbose than `-vvvvv` or so. But it seems that the bsf verbosity can only be changed via the CMake variables :/

Any ideas, or should I just remove that?

ataulien: Check Out the location where `REGOTH_LOG` is defined. We're free to control the verbosity of our logs from there.

charlydelta: I think this will take some other thinks to tweak, and this PR is already quite large. I will postpone this to another PR, together with the version output, if that's okay for you?

ataulien: Sure!

charlydelta:
@ataulien This is ready now in my book. Feel free to let me know if there's anything missing / worth revisiting / etc. :)

PR #58, opened 2019-07-19
CMake, clang warning stuff and consistency

KirmesBude:
clang was putting out warning about unused option -rdynamic for compilation. From what I gathered it is only actually needed when linking. Needs 3.13, though. 
 
I looked at the compile warnings clang reported, and found the one instance were override should have been specified. 
 
Also renamed the lonely *.h header to *.hpp for consistency.


KirmesBude:
Ohh the ci is not set up with cmake 3.13 lol.

ataulien:
Seems like the Windows build is fine. Kitware has an official way of installing the latest CMake version on Ubuntu 16.04. You could just shove that into the yaml file for linux: https://apt.kitware.com

KirmesBude:
I was not quite sure how that would work with the cmake installation already present on the image. 
Lets see.

ataulien:
you can `sudo apt-get purge cmake` before. Although updating should also work just fine.

ataulien:
Hmm, it looks like it's installing cmake 3.15 properly but then still uses the old installation?

KirmesBude:
Seems like CMake is installed manually. So I dont think purging would work. 
Maybe I am just going to remove it manually.

ataulien:
Thanks for your work, I appreciate it!

PR #59, opened 2019-07-26
Update README.md

bezxyw:
Add instruction for Fedora. Tested on Fedora 30. Had to install libXrandr-devel.


ataulien:
If that library is always required, why not add it to the other `sudo dnf ...` command?

bezxyw:
@ataulien I'm not sure if it's always required, was on my box. 

Might do some testing later on on my Centos server we'll see if it builds there without it.

ataulien:
Looks good, thanks for the PR!

PR #60, opened 2019-07-30
Inventory

ataulien:
This PR adds an inventory for the characters and a simple UI. It also implements some of the needed script externals.

For testing, a few items are given to the player in `REGothWorldViewer`. They should show up like this:


src/components/UIInventory.cpp
```diff
@@ -0,0 +1,219 @@
...
+  }
+
+  void UIInventory::setViewedInventory(HInventory inventory)
+  {
+    if (mViewedInventory)
+    {
+      // FIXME: There might be other code registering a callback here, but for some reason
+      //        we can only clear the whole thing! This should be handled differently, but
+      //        I don't know how yet.
+      mViewedInventory->OnItemChanged.clear();
```
KirmesBude: connect() returns an HEvent you can save somewhere and then call disconnect() on it. 
That should only invalidate this specific connection, right?

src/scripting/daedalus/DaedalusVMForGameWorld.cpp
```diff
@@ -447,8 +448,6 @@ namespace REGoth
       registerExternal("WLD_INSERTITEM", (externalCallback)&This::external_WLD_InsertItem);
       registerExternal("NPC_SETTALENTSKILL", (externalCallback)&This::external_NPC_SetTalentSkill);
       registerExternal("EQUIPITEM", (externalCallback)&This::external_NPC_EquipItem);
```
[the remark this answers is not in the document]
ataulien: We should probably implement the externals somewhere else or move some of the low level stuff into the parent class.

src/scripting/daedalus/DaedalusVMForGameWorld.cpp
```diff
@@ -1052,6 +1035,99 @@ namespace REGoth
...
+
+    void DaedalusVMForGameWorld::external_Npc_RemoveInvItems()
+    {
+      bs::INT32 count      = popIntValue();
+      bs::INT32 instance   = popIntValue();
+      HCharacter character = popCharacterInstance();
+
+      const bs::String& instanceName = scriptSymbols().getSymbolName(instance);
+
+      auto inventory = character->SO()->getComponent<Inventory>();
```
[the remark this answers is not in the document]
ataulien: I had the idea of making `popCharacterInstance()` a template where you can specify the part of the character you want. You need to cross the bridge from SceneObject to Component *somewhere*, so `getComponent` will be involved somehow.

src/main_WorldViewer.cpp
```diff
@@ -84,6 +86,19 @@ class REGothWorldViewer : public REGoth::REGothEngine
     mThirdPersonCamera->follow(hero);
 
     REGoth::GameplayUI::createGlobal(mMainCamera);
+
+    auto inventory = hero->SO()->getComponent<Inventory>();
+
+    inventory->giveItem("ITFOAPPLE");
```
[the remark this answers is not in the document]
ataulien: I wanted to test whether the callback fires correctly. No other reason really.

src/components/GameplayUI.hpp
```diff
@@ -96,13 +99,18 @@ namespace REGoth
...
+      return mInventory;
+    }
 
+  protected:
     void onInitialized() override;
 
     HUIDialogueChoice mChoices;
     HUISubtitleBox mSubtitleBox;
     HUIFocusText mFocusText;
+    HUIInventory mInventory;
```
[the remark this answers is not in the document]
ataulien: Renamed.

src/components/Inventory.cpp
```diff
@@ -0,0 +1,105 @@
...
+
+    OnItemChanged(instance);
+  }
+
+  const bs::Map<bs::String, bs::UINT32>& Inventory::allItems() const
+  {
+    return mInventory;
+  }
+
+  HInventory Inventory::thisHandle() const
```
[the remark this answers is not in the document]
ataulien: Ooops, I worked in the wrong file there and got confused.

src/components/UIInventory.cpp
ataulien: That's exactly what I needed.

src/components/Inventory.cpp
```diff
@@ -0,0 +1,100 @@
...
+  }
+
+  void Inventory::giveItem(const bs::String& instance, bs::UINT32 count)
+  {
+    throwIfNotUpperCase(instance);
+
+    REGOTH_LOG(Info, Uncategorized, "[Inventory] Add {0}x item {1} to Inventory of {2}", count,
+               instance, SO()->getName());
+
+    mItemCountsByInstance[instance] += 1;
```
KirmesBude: this should read count. 
What happens if we give a count of 0? This has the potential to introduce and entry with the value of 0 which should not exist in mItemCountsByInstance.

src/components/UIInventory.cpp
```diff
@@ -0,0 +1,227 @@
...
+    {
+      addItem(instance, instance, countInventory - countUI);
+    }
+    else if (countUI > countInventory)
+    {
+      removeItem(instance, countUI - countInventory);
+    }
+    else
+    {
+      // Count is the same, do nothing
```
KirmesBude: Any specific cases were this can happen?
This would only happen when calling giveItem or removeItem with count of 0.

src/components/UIInventory.cpp
```diff
@@ -0,0 +1,227 @@
...
+    if (it == mItemsByInstance.end())
+    {
+      auto& item = mItemsByInstance[instance];
+
+      item.count = count;
+      item.name  = name;
+
+      createUIElementsForItem(item);
+
+      updateItemThumbnail(item);
```
KirmesBude: Is there a usecase where you would create an UIElement for the item but not update its Thumbnail? 
If not maybe we should move it into the creation of the ui element.

src/components/UIInventory.cpp
```diff
@@ -0,0 +1,227 @@
...
+
+  void UIInventory::addItem(const bs::String& instance, const bs::String& name, bs::UINT32 count)
+  {
+    auto it = mItemsByInstance.find(instance);
+
+    if (it == mItemsByInstance.end())
+    {
+      auto& item = mItemsByInstance[instance];
+
+      item.count = count;
```
KirmesBude: What if 0 is supplied as count to this function?

src/components/GameplayUI.cpp
```diff
@@ -42,6 +42,11 @@ namespace REGoth
     {
       mFocusText = addChildElement<UIFocusText>("UIFocusText");
     }
+
+    if (!mInventory)
+    {
+      mInventory = addChildElement<UIInventory>("UIInventory");
```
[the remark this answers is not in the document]
ataulien: That'll be probably cleaner, I agree.

src/components/GameplayUI.hpp
```diff
@@ -96,13 +99,18 @@ namespace REGoth
       return mFocusText;
     }
 
-  protected:
+    HUIInventory inventory() const
```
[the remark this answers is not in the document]
ataulien: I will rename it

src/components/UIInventory.cpp
ataulien: That could also mean that the items name or value was changed as this is a general "something happened to that item" callback. Right now the only usecase is to update the count, but there might be more in the future.

Adding the else here is just defensive programming.

src/components/Inventory.cpp
ataulien: You're right! I'll take care of that!

src/components/UIInventory.cpp
```diff
@@ -0,0 +1,227 @@
...
+                                                            bs::ScrollBarType::NeverShow);
+  }
+
+  UIInventory::~UIInventory()
+  {
+  }
+
+  void UIInventory::setViewedInventory(HInventory inventory)
+  {
+    if (mViewedInventory)
```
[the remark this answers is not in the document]
ataulien: Nope, it doesn't.

src/components/UIInventory.cpp
ataulien: Should be checked, I agree.

src/components/UIInventory.cpp
ataulien: I kept those separated because then I feel like I'd have to call that method `createUIElementsForItemAndFillThem` or something. `createUIElementsForItem` is supposed to *only create* it, not access any script data, create textures, etc.

src/components/UIInventory.cpp
KirmesBude: Makes sense. I only thought about what happens right now.

PR #61, opened 2019-07-31
Refactoring: Split monolithic class, other minor refactorings

charlydelta:
With this PR I propose some changes:

- Rename `REGoth::REGothEngine` to `REGoth::Engine` and `REGoth::REGothEngineDefaultConfig` to `REGoth::EmptyGame`: The "REGoth" prefix doesn't make much sense if the namespace clearly features its affiliation, and the default config implementation was even clunkier. Now it's more clear: The engine is abstract, and the empty game is the default go-to engine.
- Move `REGoth::EngineConfig` class definition into own class
- Move unrelated functions into own file
- Group all new files into the `core` module
- Add lots of documentation, revisit and adjust old documentations
- Some clang-format stuff
- Port executables


src/CMakeLists.txt
```diff
@@ -62,8 +62,8 @@ add_library(REGothEngine STATIC
   components/Sky.hpp
   components/Sky.cpp
   core.hpp
-  core/AbstractEngine.hpp
-  core/AbstractEngine.cpp
+  core/EmptyEngine.hpp
+  core/EmptyEngine.cpp
```
ataulien: I think `EmptyGame` would be more appropriate, that was my initial suggestion too

charlydelta: Fixed now with 6679d33. Excuse me, I wasn't paying attention yesterday :(

ataulien:
I think you forgot the docs on that last one!

PR #81, opened 2019-08-05
Feature: Use proper acceleration while falling

ataulien:
Also fix "Flying" animations being affected by gravity.

Can be easily tested by throwing the nameless from a ledge in the oldmine.


src/components/CharacterAI.cpp
```diff
@@ -369,17 +378,30 @@ namespace REGoth
       rootMotion *= -1.0;
     }
 
-    if (!isStandingOnSolidGround || rootMotion.squaredLength() > 0)
+    if (isInAir || !isStandingOnSolidGround || rootMotion.squaredLength() > 0)
     {
       // Note: Gravity is acceleration, but since the walker doesn't support falling, just apply it
```
mworchel: This comment is obsolete now

src/components/CharacterAI.cpp
```diff
@@ -408,6 +428,24 @@ namespace REGoth
...
+    if (!isInAir)
+    {
+      mFallingVelocity = 0.0f;
+    }
+    else
+    {
+      mFallingVelocity += FALLING_ACCELERATION_Y * bs::gTime().getFixedFrameDelta();
+    }
+
+    if (mVisual->isPlayingFlyingAnimation())
```
mworchel: Shouldn't this be at the beginning of the function or could something go wrong? Other than falling velocity being applied in the same frame? It's a bit confusing to have this at the end. Maybe it also feels confusing as this `if` is not part of *handling* falling but it actually sets the flag which indicate how the falling is handled. Difficult to explain :smile:

src/components/CharacterAI.hpp
```diff
@@ -281,15 +288,22 @@ namespace REGoth
     // Whether Physics is being processed for this character
     bool mIsPhysicsActive = true;
 
-    // Whether this character is standing on solid ground, like the world mesh or a static vob
+    // Whether this character is standing on solid ground, like the world mesh or a static vob.
+    // Dynamic objects such as movers or other characters do not count as solid.
     bool isStandingOnSolidGround = false;
```
mworchel: Rename to `mIsStandingOnSolidGround`?

src/components/CharacterAI.hpp
```diff
@@ -281,15 +288,22 @@ namespace REGoth
...
 
+    // Whether this character is currently in air, not standing on anything.
+    bool isInAir = false;
```
mworchel: Rename to `mIsInAir`?

src/components/CharacterAI.cpp
ataulien: This is rather an overwrite to let the animation itself take care of the gravity. If I had put at the beginning of the function `FALLING_ACCELERATION_Y` would be added one time, which is not correct. I should probably just rename the function to `handleFallingAndFlying` or so. Do you have a better name? :smile:

charlydelta: `handleAirborne`?

src/components/CharacterAI.hpp
ataulien: Ooops, didn't notice that!

src/components/CharacterAI.hpp
charlydelta: or `mIsAirborne`, see the other suggestion

src/components/CharacterAI.cpp
mworchel: Oh so it's like two different states `flying` and `falling`? When would you be flying in Gothic? Hm... `handleFallingAndFlying` is a better name :smile:. Is there a way to condense the three `if`s that depend on `isInAir` any further? I find the conditional dependencies (also how `isInAir` differs from `!isStandingOnSolidGround`) still a bit confusing.

ataulien: I added some more information to `isStandingOnSolidGround` explaining what it is. Standing ontop of another dynamic object, such as a another character, is *not* solid ground.

`mIsInAir` just means that the character is not standing on anything. Maybe we can condense those into some functions in a refactoring...

> When would you be flying in Gothic?

Bloodflys do. Also during Jumping, apparently :wink:

src/components/CharacterAI.hpp
```diff
@@ -226,6 +226,20 @@ namespace REGoth
...
+     * If the character is not standing on solid ground, we need to increase
+     * the falling velocity. Once it hits the ground again, the falling velocity
+     * needs to be reset. Also some animations exist which want to take care of the falling
+     * velocity on their own.
+     */
+    void handleFallingAndFlying();
+
+    /**
+     * @return True, if the physical Character Controller should be moved or have gravity
+     *         applyied in this update cycle.
```
mworchel: applied :p

ataulien: Dammit! :smile:

PR #82, opened 2019-08-05
Feature/console basic

KirmesBude:
basics for console WIP. 
I removed the tries for now. I wont have a lot of time until september so I want to get the basics in at least - could be useful for implementing other features. 
See feature/console for another approach. 
 
Not sure I am happy with Console calling into UIConsole, because it means Console will need to be added last. But looking at "toggle frame"/etc. maybe it needs to anyways. 
But it leads to way cleaner command declarations for one.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -23,6 +23,7 @@ add_library(REGothEngine STATIC
   RTTI/RTTI_Character.hpp
   RTTI/RTTI_CharacterAI.hpp
   RTTI/RTTI_CharacterKeyboardInput.hpp
+  RTTI/RTTI_Console.hpp
   RTTI/RTTI_GameClock.hpp
   RTTI/RTTI_Inventory.hpp
   RTTI/RTTI_NodeVisuals.hpp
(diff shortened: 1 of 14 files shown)


src/CMakeLists.txt
```diff
@@ -31,8 +32,13 @@ add_library(REGothEngine STATIC
   RTTI/RTTI_ScriptObjectStorage.hpp
   RTTI/RTTI_StoryInformation.hpp
   RTTI/RTTI_TypeIDs.hpp
+  RTTI/RTTI_Console.hpp
```
charlydelta: You got that twice (line 26), remove this occurrence

src/RTTI/RTTI_Console.hpp
```diff
@@ -0,0 +1,20 @@
...
+    BS_END_RTTI_MEMBERS
+
+  public:
+    RTTI_Console()
+    {
+    }
+
+    REGOTH_IMPLEMENT_RTTI_CLASS_FOR_COMPONENT(Console)
+  };
+}  // namespace REGoth
```
charlydelta: No new line at EOF

src/RTTI/RTTI_UIConsole.hpp
```diff
@@ -0,0 +1,20 @@
...
+    BS_END_RTTI_MEMBERS
+
+  public:
+    RTTI_UIConsole()
+    {
+    }
+
+    REGOTH_IMPLEMENT_RTTI_CLASS_FOR_COMPONENT(UIConsole)
+  };
+}  // namespace REGoth
```
charlydelta: \n

mworchel:
Didn't look at all the details yet but is there a particular reason why the ´Console´ should be a ´bs::Component´? It is not serialized and not updates every frame. Why not make it a plain class that is *not even* ´bs::IReflectable´?

KirmesBude:
> Didn't look at all the details yet but is there a particular reason why the `Console` should be a `bs::Component`? It is not serialized and not updated every frame. Why not make it a plain class that is _not even_ `bs::IReflectable`?

No reason at all. 
I initially thought there might be some merit in serializing the history or datastructures for auto-completion, so I started with a component.

ataulien:
Finally got to test this! It is sooo cool to just type `insert WOLF` and have a wolf run around the oldmine! 😄

Issue #76, opened 2019-08-04
Load Sky Textures


ataulien:
The `SkyColoring`-Module also has information about the textures to be displayed in the sky, which it should pass to the related sky rendering components.  It was ported from the old REGoth repository and thus contains some blocks of code commented out regarding textures:

https://github.com/REGoth-project/REGoth-bs/blob/master/src/sky/SkyColoring.cpp

To load the textures, see the `OriginalGameResources`-module.

ataulien:
Turns out some of the sky textures are uncompressed DDS files, which is not supported by the importer yet...

ataulien:
@charlydelta Loading the sky textures has been implemented inside https://github.com/REGoth-project/BsZenLib/commit/beb53c37c6f0c5c8e478590ee66100b11575edf2.

Issue #74, opened 2019-08-04
Render Sky Dome


ataulien:
Gothic 2's sky rendering is a bit more sophisticated that what Gothic 1 did: Instead of rendering a plane, it uses a Skydome and an additional layer of clouds. This has been implemented in REGoth before, so the old implementation can serve as a reference:

https://github.com/REGoth-project/REGoth/blob/master/src/render/SkyRendering.cpp

The dome-mesh can be found in the games files if I remember correctly

Issue #73, opened 2019-08-04
Render Sky Plane


ataulien:
In Gothic 1, the sky is rendered as a plane. This has been implemented in REGoth before, so the old implementation can serve as a reference:

https://github.com/REGoth-project/REGoth/blob/master/src/render/SkyRendering.cpp

In the new implementation, I suggest creating a component which takes care of rendering the sky. A plane-mesh is available inside bsf and can be retrieved like this:

```cpp
  bs::HMesh planeMesh = bs::gBuiltinResources().getMesh(bs::BuiltinMesh::Quad);
```

PR #86, opened 2019-08-09
Refactoring: Overhaul internal Sky API, allow configuring sky mode, terminal options

charlydelta:
This MR includes:

- Introduce sky render mode
- Make sky render mode configurable
- Refactor and re-define internal sky API
- Better grouping of terminal options

This MR aims towards implementing a sky plane and sky dome by defining the basic interface.

Relevant for #73, #74, #76, but doesn't solve any of these!


src/components/SkyStateGenerator.cpp
```diff
@@ -0,0 +1,316 @@
...
+  // FIXME: There is some stuff about levelchanges here. Like, "turn off rain when on dragonisland"
+
+  // FIXME: Multiply fogColor with 0.8 when using dome!
+
+  return mCurrentSkyState;
+}
+
+std::tuple<unsigned int, unsigned int> SkyStateGenerator::findPresetIndices() const
+{
+  unsigned int p1i = static_cast<unsigned int>(mSkyStatePresets.size()) - 1;
```
ataulien: Please always use fixed width integers, such as `bs::UINT32` or `uint32_t`.

src/components/SkyStateGenerator.cpp
```diff
@@ -183,12 +183,12 @@ const std::shared_ptr<SkyState> SkyStateGenerator::update(float dayRatio)
   return mCurrentSkyState;
 }
 
-std::tuple<unsigned int, unsigned int> SkyStateGenerator::findPresetIndices() const
+std::tuple<std::uint32_t, std::uint32_t> SkyStateGenerator::findPresetIndices() const
```
ataulien: I would prefer the fixed size types provided by bsf to stay consistent actually, ie. `bs::UINT32`. Do you mind switching the code them?

src/components/SkyStateGenerator.cpp
```diff
@@ -183,12 +183,12 @@ const std::shared_ptr<SkyState> SkyStateGenerator::update(float dayRatio)
...
 {
-  unsigned int p1i = static_cast<unsigned int>(mSkyStatePresets.size()) - 1;
-  unsigned int p2i = 0;
+  std::uint32_t p1i = static_cast<std::uint32_t>(mSkyStatePresets.size()) - 1;
+  std::uint32_t p2i = 0;
 
-  for (unsigned int i = 0; i < mSkyStatePresets.size(); ++i)
+  for (std::uint32_t i = 0; i < mSkyStatePresets.size(); ++i)
```
ataulien: I think this needs a cast, otherwise there'll likely be a warning during compilation. For these simple casts, a C-Style cast is fine IMHO.

src/components/SkyStateGenerator.cpp
```diff
@@ -183,12 +183,12 @@ const std::shared_ptr<SkyState> SkyStateGenerator::update(float dayRatio)
   return mCurrentSkyState;
 }
 
-std::tuple<unsigned int, unsigned int> SkyStateGenerator::findPresetIndices() const
+std::tuple<std::uint32_t, std::uint32_t> SkyStateGenerator::findPresetIndices() const
 {
-  unsigned int p1i = static_cast<unsigned int>(mSkyStatePresets.size()) - 1;
-  unsigned int p2i = 0;
+  std::uint32_t p1i = static_cast<std::uint32_t>(mSkyStatePresets.size()) - 1;
```
ataulien: Shouldn't the `- 1` be part of the cast? Here, a simple C-Style cast is fine too, IMO

src/components/SkyStateGenerator.cpp
charlydelta: Of course not, will do. I personally just prefer to use `std` stuff, and apart from that I think I remember saying someone that bsf are thinking about deprecating those typedefs anyway (I have no source for that tho)

src/components/SkyStateGenerator.cpp
charlydelta: I always look for compilation warnings with clang, and there was none. In fact I removed a bunch of compilation warnings ;) (except for one, where I added a todo instead). Should I still cast it?

src/components/SkyStateGenerator.cpp
charlydelta: I pulled it in the cast, but I personally never use C stuff when I write C++. I can still change it if you want?

mworchel: > Here, a simple C-Style cast is fine too, IMO

Or use
```
std::uint32_t(mSkyStatePresets.size() - 1)
```
which is as expressive as the C-style cast but not C.

src/components/SkyStateGenerator.cpp
ataulien: I wonder why that is. `mSkyStatePresets.size()` should return a number of type `size_t`, which is defined to be an integer large enough to hold the entire address space, ie. 64 bit on a 64 bit machine. That said, `uint32_t` is certainly not a 64-bit number.

mworchel: Hm I don't see the reason this should produce a warning. Both are unsigned and `i` is probably just casted to the wider type for the comparison?!  Which would result in no possible loss of information.

src/components/SkyStateGenerator.cpp
ataulien: IMO, C++ style casts don't help in making the code more readable when casting between plain integer types. In fact, I mean, they are more verbose for a good reason, like when you're casting between pointer types or (god beware) casting away a const. They also solve the issue of `(uint32_t)value - 1` not being clear, which is nice.

But in this particular case we really just want to say, *Look, I have that number and it might be 64 bit but I want it to be 32*. There is no "order of operations", no pointer trickery, no consts. 

Can't we just use `auto` or `size_t` here to get rid of the casts all together?

mworchel: I mean `std::uint32_t(mSkyStatePresets.size() - 1)` is C++ and is as expressive as `(std::uint32_t)(mSkyStatePresets.size() - 1)`

src/components/SkyStateGenerator.cpp
charlydelta: Yeah this is really strange... Fixed now

src/components/SkyStateGenerator.cpp
charlydelta: @mworchel This is a function style cast and is [exactly the same thing](https://stackoverflow.com/a/3487099) as a C-style cast

mworchel: Yes, I know but it's not C syntax. Whatever.

ataulien: Oh, for some reason I completely missed this discussion, sorry for that!

I didn't even know those "function style casts" worked for numbers, this is awesome!

PR #87, opened 2019-08-11
Feature: Add macros and exception for build type aware assertion functionality

charlydelta:
This adds assertion functionality to REGoth. For build types `Debug` and `DebWithRelInfo`, assertions will throw a `REGoth::AssertionException`. For other build types, only an error message will be logged without interrupting the program.

Usage:

For any assertions:

 `REGOTH_ASSERT(myExpression() == false, "myExpression() must be false!")`;

For variable values specifically:

`REGOTH_ASSERT_VAL(myVariable, 2 <= myVariable && myVariable < 10, "Must be in range [2, 10)");`

Possible output:

```
[FATAL] A fatal error occurred and the program has to terminate!
  - Error: REGoth::AssertionException
  - Description: Assertion for 'myVariable' (= 1000) failed:  Must be in range [2, 10).  Condition: 'myVariable, 2 <= myVariable && myVariable < 10'
  - In function: virtual void REGoth::MyComponent::myFunction()
  - In file: /home/chris/projects/regoth/src/components/MyComponent.cpp:40
```

The macros support variadic parameters for `bs::StringUtil::format`, and this PR also adds support for `REGOTH_THROW`.


CMakeLists.txt
```diff
@@ -85,6 +85,11 @@ set(CXXOPTS_BUILD_TESTS OFF CACHE BOOL "" FORCE) # Disable unnecessary targets f
 add_definitions(-DCXXOPTS_NO_RTTI) # Disable RTTI dependent instructions.
 add_subdirectory(lib/cxxopts)
 
+# If CMake build type is either Debug or RelWithDebInfo, enable assertions.
```
ataulien: I think the code is expressive enough, no need to repeat it in the comment.

CMakeLists.txt
```diff
@@ -85,6 +85,11 @@ set(CXXOPTS_BUILD_TESTS OFF CACHE BOOL "" FORCE) # Disable unnecessary targets f
...
+IF(CMAKE_BUILD_TYPE MATCHES Debug OR CMAKE_BUILD_TYPE MATCHES RelWithDebInfo)
```
ataulien: Please stick to lowercase `if`, `endif` and `or`.

src/exception/Assert.hpp
```diff
@@ -0,0 +1,59 @@
...
+  do                                                                                      \
+  {                                                                                       \
+    if (!(expr))                                                                          \
+    {                                                                                     \
+      bs::String message = REGOTH_INTERNAL_ASSERT_VAL_MSG(var, expr, msg, ##__VA_ARGS__); \
+      REGOTH_THROW(REGoth::AssertionException, message);                                  \
+    }                                                                                     \
+  } while (false)
+// Disables assertions for non-debug builds.
+#else
```
ataulien: How about this?

```cpp
#else /* REGOTH_ENABLE_ASSERTIONS == false */
```

Those comments help a lot for those long ifdefs.

src/exception/Assert.hpp
```diff
@@ -0,0 +1,59 @@
+#pragma once
+
+#include <String/BsString.h>
+
+#include <exception/AssertionException.hpp>
+#include <exception/Throw.hpp>
+#include <log/logging.hpp>
+
```
ataulien: Can you include some usage examples in a comment here? Also the info that assertions might be transformed into error-logs in some build configurations would fit in nicely here.

(Don't explicitly say which build configs are affected, otherwise it's just another place to fix if we ever change that)

charlydelta:
@ataulien I think I addressed your suggestions, let me know what you think

ataulien:
Looks good to me, thank you!

PR #89, opened 2019-09-08
Performance Tweaking

ataulien:
This PR includes some changes which improve the performance of

 - Caching
 - Gameplay

The two main changes are:

 - Disable physics on all characters by default
 - Call characters event queue module less often
 - Cache resources in bulk (Multithreaded), before actually loading the ZEN file


src/core/Engine.cpp
```diff
@@ -22,6 +22,7 @@
 #include <log/logging.hpp>
 #include <original-content/OriginalGameFiles.hpp>
 #include <original-content/VirtualFileSystem.hpp>
+#include "original-content/OriginalGameResources.hpp"
```
charlydelta: I would stick with one include style. I personally prefer `<>`. The includes are also not sorted alphabetically (might be because of `""` vs `<>`)

src/main_WorldCacheTest.cpp
```diff
@@ -2,64 +2,27 @@
...
 #include <core.hpp>
-#include <components/AnchoredTextLabels.hpp>
-#include <components/GameWorld.hpp>
-#include <components/Item.hpp>
-#include <components/Waynet.hpp>
-#include <components/Waypoint.hpp>
 #include <exception/Throw.hpp>
 #include <log/logging.hpp>
 #include <original-content/VirtualFileSystem.hpp>
+#include <BsZenLib/CacheUtility.hpp>
```
charlydelta: Again a small include style nitpick: `BsZenLib` is a dependency, and not the code of this repo. Therefore, it should be in a separate block above this repository's code:

(At least I do it like that and think it is useful, and `clang-format` works pretty good with that scheme. What's your thoughts on stuff like that?)

```
#include <memory>

#include <BsFPSCamera.h>
#include <Components/BsCCamera.h>
#include <Utility/BsTimer.h>

#include <BsZenLib/CacheUtility.hpp>

#include <core.hpp>
#include <exception/Throw.hpp>
#include <log/logging.hpp>
#include <original-content/VirtualFileSystem.hpp>
```

src/components/CharacterEventQueue.cpp
```diff
@@ -305,7 +305,7 @@ namespace REGoth
         break;
 
       case AI::StateMessage::ST_Wait:
-        message.waitTime -= bs::gTime().getFixedFrameDelta();
+        message.waitTime -= bs::gTime().getFrameDelta();
```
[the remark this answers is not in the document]
ataulien: If you're running with VSync, the dynamic one will still be capped to 60 fps. Since my laptop struggles to reach 60 fps (integrated graphics), not using fixed-update increases the frame-rate by a lot.

src/core/Engine.cpp
ataulien: Right, no idea how that slipped in!

src/main_WorldCacheTest.cpp
ataulien: I think it's a nice way of organizing includes. I just have to get used to it!

ataulien:
> I just don't entirely grasp how using update instead of fixedUpdate improves the performance?

If you're running with VSync, the dynamic update will still be capped to 60 fps. Since my laptop struggles to reach 60 fps (integrated graphics), not using fixed-update increases the frame-rate by a lot.

charlydelta:
Ah all right, I see

src/main_WorldCacheTest.cpp
charlydelta: Great :+1:

Issue #113, opened 2019-11-13
Crash "NotCoreThead"


KirmesBude:
```
[FATAL] A fatal error occurred and the program has to terminate!
  - Error: InternalErrorException
  - Description: This method can only be accessed from the core thread.
  - In function: void bs::throwIfNotCoreThread()
  - In file: /home/bude/myGit/REGoth-bs/lib/bsf/Source/Foundation/bsfCore/CoreThread/BsCoreThread.cpp:384

Stack trace: 
0) lib/libbsf.so.1: bs::CrashHandler::getStackTrace[abi:cxx11]()+0x1de [0x7f3d1a3573fe]
1) lib/libbsf.so.1: bs::CrashHandler::logErrorAndStackTrace(std::__cxx11::basic_string<char, std::char_traits<char>, bs::StdAlloc<char, bs::GenAlloc> > const&, std::__cxx11::basic_string<char, std::char_traits<char>, bs::StdAlloc<char, bs::GenAlloc> > const&, std::__cxx11::basic_string<char, std::char_traits<char>, bs::StdAlloc<char, bs::GenAlloc> > const&, std::__cxx11::basic_string<char, std::char_traits<char>, bs::StdAlloc<char, bs::GenAlloc> > const&, unsigned int) const+0x354 [0x7f3d1a334444]
2) lib/libbsf.so.1: bs::CrashHandler::reportCrash(std::__cxx11::basic_string<char, std::char_traits<char>, bs::StdAlloc<char, bs::GenAlloc> > const&, std::__cxx11::basic_string<char, std::char_traits<char>, bs::StdAlloc<char, bs::GenAlloc> > const&, std::__cxx11::basic_string<char, std::char_traits<char>, bs::StdAlloc<char, bs::GenAlloc> > const&, std::__cxx11::basic_string<char, std::char_traits<char>, bs::StdAlloc<char, bs::GenAlloc> > const&, unsigned int) const+0x5f [0x7f3d1a357d2f]
3) lib/libbsf.so.1: bs::throwIfNotCoreThread()+0x22a [0x7f3d1a37a77a]
4) lib/libbsf.so.1: bs::ct::CoreObject::~CoreObject()+0x17 [0x7f3d1a37ebd7]
[...]

PR #114, opened 2019-11-17
Stop destroying newly cloned materials.

crushack:
Note: this lead to a crash referenced in:
https://github.com/REGoth-project/REGoth-bs/issues/113


ataulien:
I'll merge this so people can get REGoth running again, but I'll leave #113 open until we figured out whether this results in a memory leak. Thanks for the fix!

PR #119, opened 2020-03-07
Feature/GitHub actions

KirmesBude:
Does not cache between branches, so every new PR needs ~20-30 minutes for its first run. (Actually apparently if there is a cache on master, every branch based off of that will use the master cache) 
Also not sure what happens if you update bsf on a branch that already has a cache, since the step is fully skipped - I guess that is fixed by including the submodule commit id in the cache key in the latest commit.


KirmesBude:
Should not have triggered a new build when doing the README changes. 
Possibly because the other build was still running?

ataulien:
That's a great addition, thanks for taking the time to fiddle with the CI! I tried once and failed 😄

PR #122, opened 2020-03-13
macOS build support

nikitamiroshnichenko:


src/CMakeLists.txt
```diff
@@ -178,6 +178,13 @@ add_library(REGothEngine STATIC
   world/internals/ImportSingleVob.hpp
   )
 
+# link necessary libraries for physfs on macOS
```
ataulien: This says "physfs", is that a copy-paste mistake?

nikitamiroshnichenko: linkage errors coming from PhysicsFS library. So yes, may be this should be "PhysicsFS" instead of  "physfs". Reason why I typed like that because PhysicsFS symbols looks like this:"___PHYSFS_platformCalcBaseDir"

ataulien: Oh, I thought the libraries were needed by REGothEngine itself. "PhysFS" is fine, "PhysicsFS" would be wrong.

